PR #21, opened 2019-09-18
ad9361: sample rate setter

777arc:
updated sample rate setter to match behavior in ad9361_baseband_auto_rate.c.  Allows Pluto to go down to 521kHz (couldn't go lower than around 2.5MHz prior).  I tested it by tuning to the FM band and "zooming" in and out, and it seemed to work fine, but it wasn't a comprehensive test.

### DIFF
--- a/adi/ad9361.py
+++ b/adi/ad9361.py
@@ -131,8 +131,66 @@ class ad9361(rx_tx, context_manager):
         return self._get_iio_attr("voltage0", "sampling_frequency", False)
 
     @sample_rate.setter
-    def sample_rate(self, value):
-        self._set_iio_attr("voltage0", "sampling_frequency", False, value)
+    def sample_rate(self, rate):
+        if (rate < 521e3):
+            raise ValueError("Error: Does not currently support sample rates below 521e3")
+
+        # The following was converted from ad9361_set_bb_rate() in libad9361-iio
+        if (rate <= 20000000):
+            dec = 4
+            fir = [-15,-27,-23,-6,17,33,31,9,-23,-47,-45,-13,34,69,67,21,-49,-102,-99,-32,69,146,143,48,-96,-204,-200,-69,129,278,275,97,-170,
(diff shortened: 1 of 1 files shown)


adi/ad9361.py
```diff
@@ -131,8 +131,68 @@ def sample_rate(self):
...
+            max_rate = (dacrate // txrate) * 16
+            if (max_rate < taps):
+                self._set_iio_attr("voltage0", "sampling_frequency", False, 3000000)
+            self._set_iio_attr("out", "voltage_filter_fir_en", False, 1)
+            self._set_iio_attr("voltage0", "sampling_frequency", False, rate)
+        else:
+            self._set_iio_attr("voltage0", "sampling_frequency", False, rate)
+            self._set_iio_attr("out", "voltage_filter_fir_en", False, 1) 
+            
+        return 0
```
tfcollins: Return here is not necessary

adi/ad9361.py
```diff
@@ -131,8 +131,68 @@ def sample_rate(self):
         return self._get_iio_attr("voltage0", "sampling_frequency", False)
 
     @sample_rate.setter
-    def sample_rate(self, value):
-        self._set_iio_attr("voltage0", "sampling_frequency", False, value)
+    def sample_rate(self, rate):
+        if (rate < 521e3):
+            raise SystemExit("Error: Does not currently support sample rates below 521e3")
```
tfcollins: SystemExit might be a tad too severe since it exits the interpreter. Standard exception should be enough

tfcollins:
Be sure to run the pre-commit checks `pre-commit run --all-files` and sign your commits.

777arc:
pre-commit is erroring out during it's initialization (nothing to do with your code), and I'm on a pretty fresh and standard ubuntu 18 machine.  I was able to sign the commit though, glad you pointed out that ability, never used it before.  Now time to squash the 2 into 1.

tfcollins:
@777arc Thanks for the PR

PR #32, opened 2019-11-15
Add support for AD7124

adisuciu:
The AD7124 driver still has some issues, however this should be the API.

A few notes:
- Currently it only works when one channel is enabled.
- There are sample rates for all channels (and they are not shared attributes), however i'm not sure how it is supposed to work when having multiple channels enabled - same buffer, multiple rates. This is why I set the sample rate for all channels
- The scale and scale_available attribute was trickier because the str(float val) gives a value like 4.56+e06 which is incompatible with IIO.
- I'm not sure where the to_volts method should be implemented as it uses different coeficients for different channels.
- I created the channels list on __init__ dynamically. I do this because the number of channels is reconfigurable via devicetree. Does it make sense reffering to channels by index ? Or you should refer to channels by name. A channel name for ad7124 is "input voltageX-voltageY" because the input is always differential.
- The example just gets data from the ADC. I can create some plots when the driver will work properly. It currently gives out junk values.

Signed-off-by: Adrian Suciu <adrian.suciu@analog.com>

### DIFF
--- a/adi/ad7124.py
+++ b/adi/ad7124.py
@@ -1,0 +1,116 @@
+# Copyright (C) 2019 Analog Devices, Inc.
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification,
+# are permitted provided that the following conditions are met:
+#     - Redistributions of source code must retain the above copyright
+#       notice, this list of conditions and the following disclaimer.
+#     - Redistributions in binary form must reproduce the above copyright
+#       notice, this list of conditions and the following disclaimer in
+#       the documentation and/or other materials provided with the
+#       distribution.
+#     - Neither the name of Analog Devices, Inc. nor the names of its
+#       contributors may be used to endorse or promote products derived
+#       from this software without specific prior written permission.
+#     - The use of this software may or may not infringe the patent rights
+#       of one or more patent holders.  This license does not release you
+#       from the requirement that you obtain separate licenses from these
+#       patent holders to use this software.
+#     - Use of the software either in source or binary form, must be run
+#       on or directly connected to an Analog Devices Inc. component.
+#
+# THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES,
+# INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT, MERCHANTABILITY AND FITNESS FOR A
+# PARTICULAR PURPOSE ARE DISCLAIMED.
+#
+# IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, INTELLECTUAL PROPERTY
+# RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR
+# BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+# THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from decimal import Decimal
+
+import numpy as np
+from adi.attribute import attribute
+from adi.context_manager import context_manager
+from adi.rx_tx import rx
+
+
+class ad7124(rx, context_manager):
+    """ AD7124 ADC """
+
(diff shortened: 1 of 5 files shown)


adi/ad7124.py
```diff
@@ -0,0 +1,116 @@
...
+    @property
+    def sampling_frequency(self):
+        # only works with the same frequency for all channels
+        return self._get_iio_attr(self.channel[0].name, "sampling_frequency", False)
+
+    @sampling_frequency.setter
+    def sampling_frequency(self, value):
+        for ch in self.channel:
+            self._set_iio_attr(ch.name, "sampling_frequency", False, value)
+        return 1
```
tfcollins: Decorator based setters don't return things in python

adi/ad7124.py
```diff
@@ -0,0 +1,116 @@
...
+    _device_name = ""
+
+    def __init__(self, uri=""):
+
+        context_manager.__init__(self, uri, self._device_name)
+        self._ctrl = self._ctx.find_device("ad7124-8")
+        self._rxadc = self._ctx.find_device("ad7124-8")
+
+        # dynamically get channels
+        for ch in self._ctrl._channels:
```
tfcollins: I assume this works since all channels have scan elements?

adi/ad7124.py
```diff
@@ -0,0 +1,116 @@
...
+    def sampling_frequency(self, value):
+        for ch in self.channel:
+            self._set_iio_attr(ch.name, "sampling_frequency", False, value)
+        return 1
+
+    @property
+    def scale_available(self):
+        return self._get_iio_attr(self.channel[0].name, "scale_available", False)
+
+    class _channel(attribute):
```
tfcollins: This is pretty handy to have. Do you think we could possibly combine this into the rx_tx class so data received from buffers are automatically converted to specific units? Say we had a property like rx_output_type = 'SI' or something?

The adis16460 class would benefit.

adi/ad7124.py
```diff
@@ -0,0 +1,116 @@
...
+        # dynamically get channels
+        for ch in self._ctrl._channels:
+            name = ch._id
+            self._rx_channel_names.append(name)
+            self.channel.append(self._channel(self._ctrl, name))
+        rx.__init__(self)
+
+    @property
+    def sampling_frequency(self):
+        # only works with the same frequency for all channels
```
tfcollins: Can you add docstrings to each property

examples/ad7124.py
```diff
@@ -0,0 +1,56 @@
...
+from scipy import signal
+
+# Set up AD7124
+ad7124 = adi.ad7124(uri="ip:192.168.1.171")
+
+sc = ad7124.scale_available
+ad7124.channel[0].scale = sc[2]
+
+ad7124.sampling_frequency = 19200 # sets sample rate for all channels
+ad7124.rx_enabled_channels = [ 0 ] # currently only one enabled channel buffer works at a time
```
tfcollins: Is this how to it works with IIO-Scope or maybe something is wrong with core pyadi-iio which doesn't support this case?

tfcollins:
One reason you might be getting bad data is that the buffers default to np.int16 samples and the AD7123 is a 24-bit device (I believe) which are probably put into a 32-bit container in DDR. You can update the expected type like this: https://github.com/analogdevicesinc/pyadi-iio/blob/master/adi/adis16460.py#L52

adi/ad7124.py
adisuciu: Yes. Is there any other way to do it to catch channels without buffers too ?

examples/ad7124.py
adisuciu: It's the same in IIO-oscilloscope. It looks like it's a driver thing.

adi/ad7124.py
adisuciu: added

adi/ad7124.py
adisuciu: added

adisuciu:
> One reason you might be getting bad data is that the buffers default to np.int16 samples and the AD7123 is a 24-bit device (I believe) which are probably put into a 32-bit container in DDR. You can update the expected type like this: https://github.com/analogdevicesinc/pyadi-iio/blob/master/adi/adis16460.py#L52

I changed the return type. Some of the data is valid because there is a correlation between the test signal and the plot. I get data interleaved with zeroes and some junk data as well. It's the same in IIO-Oscilloscope.

adi/ad7124.py
adisuciu: Do you think this should be a separate pull request ?

adisuciu: I added another commit that handles _rx_output_type .. seems to work for AD7124. Does the implementation make sense. I don't have a ADIS16460. I can probably test it tomorrow if I find one in the office.

tfcollins: There is one inside the PackRF

adi/ad7124.py
tfcollins: You could check if the scan_element properties of the channel objects are true and if they are inputs.

examples/ad7124.py
tfcollins: I guess we'll live with it for now. Will revisit when the driver is done/fixed

PR #40, opened 2019-12-11
(WIP) Add Lidar

mihaiolteanu:
Add device ad9094
Add device ad5627
Add Lidar which uses both devices from above and also configures 'pulse-capture'
Add Lidar GUI example

### DIFF
--- a/adi/ad5627.py
+++ b/adi/ad5627.py
@@ -1,0 +1,91 @@
+# Copyright (C) 2019 Analog Devices, Inc.
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification,
+# are permitted provided that the following conditions are met:
+#     - Redistributions of source code must retain the above copyright
+#       notice, this list of conditions and the following disclaimer.
+#     - Redistributions in binary form must reproduce the above copyright
+#       notice, this list of conditions and the following disclaimer in
+#       the documentation and/or other materials provided with the
+#       distribution.
+#     - Neither the name of Analog Devices, Inc. nor the names of its
+#       contributors may be used to endorse or promote products derived
+#       from this software without specific prior written permission.
+#     - The use of this software may or may not infringe the patent rights
+#       of one or more patent holders.  This license does not release you
+#       from the requirement that you obtain separate licenses from these
+#       patent holders to use this software.
+#     - Use of the software either in source or binary form, must be run
+#       on or directly connected to an Analog Devices Inc. component.
+#
+# THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES,
+# INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT, MERCHANTABILITY AND FITNESS FOR A
+# PARTICULAR PURPOSE ARE DISCLAIMED.
+#
+# IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, INTELLECTUAL PROPERTY
+# RIGHTS, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR
+# BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+# THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from adi.context_manager import context_manager
+from adi.rx_tx import tx
+
+class ad5627(tx, context_manager):
+    """ AD5627 Low Power Dual nanoDAC """
+
+    _complex_data     = False
+    _tx_channel_names = ["voltage0", "voltage1"]
+    _device_name      = ""
+
+    def __init__(self, uri=""):
+        context_manager.__init__(self, uri, self._device_name)
+        self._txdac = self._ctx.find_device("ad5627")
+        tx.__init__(self)
+
+    @staticmethod
+    def _adp_bias_volts_to_raw_convert(value, inverse):
+        """Convert ADP bias from user values to internal values (or reverse)."""
+        ret = 0.0
+        if (inverse):
+            ret = -((value * 5 * 18.18) / 4096) + 122
+        else:
+            ret = ((-122 - value) * 4096) / (5 * 18.18)
+        return ret
+
+    @property
+    def apdbias(self):
+        """Get the APD Bias."""
+        bias = self._get_iio_attr("voltage0", "raw", True, self._txdac)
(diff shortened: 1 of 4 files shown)


commodo:
Can this code run on Python3?
Py2 is kind-of-going away, so we should consider Py3 first, and Py2 second.

commodo:
Commits should have Signed-off-by tags

adi/ad5627.py
```diff
@@ -0,0 +1,83 @@
...
+        ret = 0.0
+        if (inverse):
+            ret = (value * 5) / 4096
+        else:
+            ret = (value * 4096) / 5
+        return ret
+
+    @property
+    def tiltvoltage(self):
+        voltage = self._get_iio_attr("voltage1", "raw", True, self._txdac)
```
tfcollins: Add docstring

adi/ad5627.py
```diff
@@ -0,0 +1,83 @@
...
+        return self._tilt_volts_to_raw_convert(voltage, True)
+
+    @tiltvoltage.setter
+    def tiltvoltage(self, voltage):
```
tfcollins: Add docstring

adi/lidar.py
```diff
@@ -0,0 +1,110 @@
...
+# BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+# THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from adi.ad5627 import ad5627
+from adi.ad9094 import ad9094
+from adi.rx_tx import phy
+import numpy as np
+
+class Lidar (ad5627, ad9094, phy):
```
tfcollins: The class name is too generic since we are likely to support multiple lidar systems in the future. Please change to eval board name

adi/lidar.py
```diff
@@ -0,0 +1,110 @@
...
+    def __init__(self, uri, pulse_capture_address="7c700000"):
+        ad5627.__init__(self, uri)
+        ad9094.__init__(self, uri)
+        # The name of the pulse capture channel. The address might be different,
+        # depending on how the board was built.
+        self.pulse_capture = pulse_capture_address + "." + "axi-pulse-capture"
+        self._ctrl = self._ctx.find_device(self.pulse_capture)
+        self.set_all_iio_attrs_to_default_values()
+
+    def read_channels(self):
```
tfcollins: To keep uniformity across the classes it might be better to simply rename this method rx, so it won't be called by mistake. What is the purpose of the zeroed out channel? Would it ever be useful for the user to have access too?

adi/lidar.py
```diff
@@ -0,0 +1,110 @@
...
+        ret = []       
+        for channel in self.rx():
+            cast_channel = channel.astype(np.int8)
+            # Ignore Channel4 and all everything else.
+            if not np.all(cast_channel==0):
+                ret.append(cast_channel)
+        return ret
+
+    def laser_enable(self):
+        self._set_iio_attr_int("altvoltage0", "en", True, 1, self._ctrl)
```
tfcollins: Add docstring

adi/lidar.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def laser_disable(self):
+        self._set_iio_attr_int("altvoltage0", "en", True, 0, self._ctrl)
+
+    def set_laser_pulse_width(self, width):
+        self._set_iio_attr_int("altvoltage0", "pulse_width_ns", True, width, self._ctrl)
```
tfcollins: Add docstring

adi/lidar.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def set_laser_frequency(self, frequency):
+        self._set_iio_attr_int("altvoltage0", "frequency", True, frequency, self._ctrl)
```
tfcollins: Add docstring

adi/lidar.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def set_channel_sequencer_enable_disable(self, status):
+        """Status = 0 to disable, 1 to enable the sequencer."""
```
tfcollins: Most other methods that address individual IIO properties are using decorators. Is there a specific reason this is not being done here as well?

mihaiolteanu:
> Can this code run on Python3?

Python3 all the way.

adi/ad5627.py
```diff
@@ -0,0 +1,83 @@
...
+        ret = 0.0
+        if (inverse):
+            ret = -((value * 5 * 18.18) / 4096) + 122
+        else:
+            ret = ((-122 - value) * 4096) / (5 * 18.18)
+        return ret
+
+    @property
+    def apdbias(self):
+        bias = self._get_iio_attr("voltage0", "raw", True, self._txdac)
```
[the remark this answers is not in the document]
mihaiolteanu: Added.

adi/ad5627.py
```diff
@@ -0,0 +1,83 @@
...
+        return self._adp_bias_volts_to_raw_convert(bias, True);
+
+    @apdbias.setter
+    def apdbias(self, value):
+        bias = self._adp_bias_volts_to_raw_convert(value, False);
```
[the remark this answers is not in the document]
mihaiolteanu: Added.

adi/ad5627.py
mihaiolteanu: Added.

adi/ad5627.py
mihaiolteanu: Added.

adi/lidar.py
mihaiolteanu: Changed to fmclidar1.

adi/lidar.py
mihaiolteanu: I think that this change will transform this function into a recursive one, since I'm calling `self.rx()` in its body.

adi/lidar.py
mihaiolteanu: Added.

adi/lidar.py
mihaiolteanu: Added.

adi/lidar.py
mihaiolteanu: Added.

adi/lidar.py
mihaiolteanu: I'm not interested in reading this attribute, only in setting it.

adi/lidar.py
mihaiolteanu: That channel is only relevant to HW and not to the user. In short,

> Would it ever be useful for the user to have access too?

No

tfcollins: Use the [super](https://docs.python.org/3/library/functions.html#super) method to call the parent class's rx.

adi/lidar.py
tfcollins: Would anyone ever want to know the state of this attribute? Also, what about the other set methods like: set_channel_sequencer_opmode, set_channel_sequencer_order_auto_mode, ...

adi/ad5627.py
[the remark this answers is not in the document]
tfcollins: I think this file got missed

adi/lidar.py
mihaiolteanu: I've changed all the `set` interfaces to properties.

adi/lidar.py
mihaiolteanu: Neat! It did the trick :)

adi/ad5627.py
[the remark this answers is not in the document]
mihaiolteanu: Everything docstring'ed

PR #43, opened 2019-12-23
Add ADXL345 support

mthoren-adi:
Add ADXL345 support
AD7124: Add type annotation to pass mypy

### DIFF
--- a/adi/adxl345.py
+++ b/adi/adxl345.py
@@ -32,91 +32,70 @@
 # THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 from decimal import Decimal
-from typing import Any
 
 import numpy as np
 from adi.attribute import attribute
 from adi.context_manager import context_manager
-from adi.rx_tx import rx
 
 
-class ad7124(rx, context_manager):
-    """ AD7124 ADC """
+class adxl345(context_manager, attribute):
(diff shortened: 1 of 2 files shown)


adi/adxl345.py
```diff
@@ -0,0 +1,101 @@
...
+            self._set_iio_attr(
+                self.name, "sampling_frequency", False, str(Decimal(value).real)
+            )
+
+        @property
+        def scale(self):
+            """ADXL345 channel scale(gain)"""
+            return float(self._get_iio_attr_str(self.name, "scale", False))
+
+    def to_g(self, ch):
```
tfcollins: This functionality might already exist: https://github.com/analogdevicesinc/pyadi-iio/blob/master/adi/rx_tx.py#L148
aka setting the rx_output_type property will get you the same thing

tfcollins:
Thanks a lot Mark :) Can you sign the commit too?

adi/adxl345.py
mthoren-adi: Looks like this is for buffered channels only? It might be handy to have a general-purpose (unbuffered) channel class that includes SI conversion.

tfcollins: Sounds reasonable. Have an idea of what the API would look like? Maybe:
`data = dev.rx_unbuffered()`
or 
`data = dev.rx('unbuffered')`
Just spitballing here

PR #58, opened 2020-01-30
Ad7124

adisuciu:
Added some data validation/truncation to the rx path
Fixed ad7124 buffered mode

### DIFF
--- a/adi/ad7124.py
+++ b/adi/ad7124.py
@@ -45,8 +45,8 @@ class ad7124(rx, context_manager):
     _complex_data = False
     channel = []  # type: ignore
     _device_name = ""
-    _rx_data_type = '>u4'
-    _rx_mask = 0xffffff
+    _rx_data_type = ">u4"
+    _rx_mask = 0xFFFFFF
     _rx_shift = 0
 
     def __init__(self, uri="", part="ad7124-8"):
@@ -66,11 +66,13 @@ class ad7124(rx, context_manager):
     def sample_rate(self):
         """Sets sampling frequency of the AD7124"""
         return self._get_iio_attr(self.channel[0].name, "sampling_frequency", False)
+        # return self._get_iio_dev_attr("sampling_frequency")
 
     @sample_rate.setter
     def sample_rate(self, value):
         for ch in self.channel:
             self._set_iio_attr(ch.name, "sampling_frequency", False, value)
+        # self._set_iio_dev_attr_str("sampling_frequency",value)
 
     @property
     def scale_available(self):
--- a/adi/ad7124.py
+++ b/adi/ad7124.py
@@ -80,13 +80,11 @@ class ad7124(rx, context_manager):
     def sample_rate(self):
         """Sets sampling frequency of the AD7124"""
         return self._get_iio_attr(self.channel[0].name, "sampling_frequency", False)
-        # return self._get_iio_dev_attr("sampling_frequency")
 
     @sample_rate.setter
     def sample_rate(self, value):
         for ch in self.channel:
             self._set_iio_attr(ch.name, "sampling_frequency", False, value)
-        # self._set_iio_dev_attr_str("sampling_frequency",value)
 
     @property
     def scale_available(self):
--- a/adi/ad7124.py
+++ b/adi/ad7124.py
@@ -76,6 +76,23 @@ class ad7124(rx, context_manager):
             self.channel.append(self._channel(self._ctrl, name))
         rx.__init__(self)
 
+    def rx(self):
+        sig = super().rx()
+
+        if (
+            self._rx_unbuffered_data
+            or self._complex_data
+            or self.rx_output_type == "raw"
+        ):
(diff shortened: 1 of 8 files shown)


tfcollins:
Failing CI. Be sure to run the precommit checks.

adisuciu:
Something is wrong with my pre-commit hooks but, fixed .. finally.

tfcollins:
Driver issues resolved?

tfcollins:
Whats the status here? Data correct now?

adisuciu:
Well, the data is correct, however the driver still hangs in some corner cases. The fix won't affect the interface hopefully, and if it does we'll do another PR. 
We just need to update part compatibility process, rebase to master, and we'll be ready for merge.

tfcollins:
Do you want me to cleanup the commits?

adisuciu:
No, just review the latest commits, and if all good, it can be merged.

adi/rx_tx.py
```diff
@@ -106,6 +109,10 @@ def __del__(self):
         self._rxadc = []
 
     def _rx_init_channels(self):
+        for m in self._rx_channel_names:
```
tfcollins: This loop will break other devices and ignores the rx_enabled_channels property

adi/rx_tx.py
```diff
@@ -210,7 +224,8 @@ def __rx_non_complex(self):
 
             for c in range(stride):
                 raw = x[c::stride]
-                sig.append(raw * rx_scale[c] + rx_offset[c])
+                # when scale is used, values are in mV
+                sig.append((raw * rx_scale[c] + rx_offset[c]) / 1000.0)
```
tfcollins: This change should be in the device specific class not the core.

examples/ad7124.py
```diff
@@ -37,22 +37,21 @@
...
 ad7124.rx_output_type = "SI"
 
-# sets sample rate for all channels
-ad7124.sample_rate = 19200
-# currently only one enabled channel buffer works at a time
-ad7124.rx_enabled_channels = [0]
+ad7124.sample_rate = 19200  # sets sample rate for all channels
+ad7124.rx_enabled_channels = [
+    ad_channel
+]  # currently only one enabled channel buffer works at a time
```
tfcollins: If this is a part limitation can you provide an overloaded rx_enabled_channel set property then?

adi/ad7124.py
```diff
@@ -64,11 +80,13 @@ def __init__(self, uri="", part="ad7124-8"):
     def sample_rate(self):
         """Sets sampling frequency of the AD7124"""
         return self._get_iio_attr(self.channel[0].name, "sampling_frequency", False)
+        # return self._get_iio_dev_attr("sampling_frequency")
 
     @sample_rate.setter
     def sample_rate(self, value):
         for ch in self.channel:
             self._set_iio_attr(ch.name, "sampling_frequency", False, value)
+        # self._set_iio_dev_attr_str("sampling_frequency",value)
```
tfcollins: Remove

adi/ad7124.py
```diff
@@ -64,11 +80,13 @@ def __init__(self, uri="", part="ad7124-8"):
     def sample_rate(self):
         """Sets sampling frequency of the AD7124"""
         return self._get_iio_attr(self.channel[0].name, "sampling_frequency", False)
+        # return self._get_iio_dev_attr("sampling_frequency")
```
tfcollins: Remove

adi/rx_tx.py
```diff
@@ -55,7 +55,10 @@ class rx(attribute):
     _complex_data = False
     _rx_data_type = np.int16
     _rx_data_si_type = np.int16
+    _rx_mask = 0x0000
+    _rx_shift = 0
     rx_output_type = "raw"
+
```
tfcollins: nip pick, random space

examples/ad7124.py
adisuciu: It used to be, now it should work with multiple channels. The performance is better/worse based on the sample rates of the channels.

adi/rx_tx.py
adisuciu: The whole SI part(lines 204-228) or just the mV part (225-228) ?

adi/rx_tx.py
adisuciu: Yes, I think it will break buffers that were already acquired but not pulled. I didn't think of that. 

However, the problem here is that if you enable a channel there is no way to disable it. Do you think we should loop over all channels at line 114/120 and do enabled true/false if the current iterator is in rx_enabled_channels ?

adi/rx_tx.py
tfcollins: 228 the divide by 1000

adi/rx_tx.py
tfcollins: We could probably add a check to see if a buffer exists then destroy it before modifying enabled channels

PR #61, opened 2020-02-05
Added new parameterized tests.

kister-jimenez:
-Added test_daq2_p.py test_ad9680_p.py and test_pluto_p.py
-Added fmcomms2 to the check_board_other to detect fmcomms2 board in test_ad9361_p test
-Added a uri property to the BoardInterface/BaseTestHelpers to let the test fixtures use the uri of the previously detected matching board.
-usb context uses the ctx (i.e. 'usb:1.2.5') if ctx.name != "local"


kister-jimenez:
FMCOMMS5 replacement is done since its the same as ad9361 with _chip_b. But I will have to do the testing first on the actual hardware to verify. I will add it in this pull request as soon as tested.

tfcollins:
The libiio bindings need to be on path for isort to work correctly. Otherwise it segregates the unknown imports

kister-jimenez:
> 
> 
> The libiio bindings need to be on path for isort to work correctly. Otherwise it segregates the unknown imports

I installed the pyadi-iio and the libiio bindings from source that's why I'm getting different ordering with isort. So my pyadi-iio and iio imports gets treated together as third-party libraries while the one in the CI iio is treated as a local library.

PR #63, opened 2020-02-07
Add parametrized test for FMComms5

kister-jimenez:
- Added parametrized test for FMComms5.
- Added cyclic buffer testing and cyclic buffer exception testing routines and fixtures in conftest.
- Fixed the multiline strings formatting in the Thrown exceptions.
- Added DDS Loopback testing routine and fixture in conftest.
- Removed old tests for fmcomms5, ad9680, and pluto.


kister-jimenez:
I have tested the new pluto test including the dds loopback test I added.

I have random fails with FMComms5 parametrized testing with the chip B. It randomly fails in sfdr, and iq_loopback tests. Sometimes it passes in these tests. When it fails in the sfdr, it only detects 1 peak so it gets an indexing Exception. For iq_loopback tests that fail, the peak_frequency is detected in the less than 1kHz instead of around 400kHz. I'm not sure yet if it's my board that has problems.

test/test_fmcomms5_p.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+#########################################
+@pytest.mark.parametrize("classname, hardware", [(classname, hardware)])
+@pytest.mark.parametrize("channel", [0])
+@pytest.mark.parametrize(
+    "param_set",
+    [
+        dict(
+            tx_lo=1000000000,
```
tfcollins: Might me useful to set the chip b LOs to somewhere away from this frequencies or you can get images from those transmitters

test/test_fmcomms5_p.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+#########################################
+@pytest.mark.parametrize("classname, hardware", [(classname, hardware)])
+@pytest.mark.parametrize("channel", [3])
+@pytest.mark.parametrize(
+    "param_set",
+    [
+        dict(
+            tx_lo_chip_b=1000000000,
```
tfcollins: Same as above for chip a

tfcollins:
Based on your comment I'm not totally sure you want this merged yet. @kister-jimenez just let me know

tfcollins:
This should help with isort formatting issues: https://github.com/analogdevicesinc/pyadi-iio/commit/b8c2a328834c519d3a7ad6c37c07d1c2abb4ed68

kister-jimenez:
> 
> 
> Based on your comment I'm not totally sure you want this merged yet. @kister-jimenez just let me know

Yes, don't merge this yet. I'll do your suggestions and test a bit on the board.

kister-jimenez:
The failures were caused by the current release candidate linux image. It works fine in 2018_R2 image. Anyway, the chip_b loopback tests are not valid since I noticed that the chip_b dac and adc channels are not used in the rx_tx and dds. Only the attributes were accessed. Should I submit this as an issue or is there someone already working on this? I'll work on it if no ones working on it yet.

tfcollins:
FMComms5 does have unified DMA drivers and the DDS should be handled with the _split_cores property (https://github.com/analogdevicesinc/pyadi-iio/blob/master/adi/dds.py#L44)
Does that not cover that functionality?

kister-jimenez:
> 
> 
> FMComms5 does have unified DMA drivers and the DDS should be handled with the _split_cores property (https://github.com/analogdevicesinc/pyadi-iio/blob/master/adi/dds.py#L44)
> Does that not cover that functionality?

I see. You're right. I was confused at first on how to enable the channels. But looking at the iio_info of fmcomms5, the buffer capable devices are only displayed in the chip A. Sorry for the confusion. Thank you.

tfcollins:
Np. FMComms5 is rather strange :) Can I merge this?

kister-jimenez:
I'll just add some DDS test routines so probably you can merge tomorrow.

kister-jimenez:
If you are fine with this, you can merge this. I left the non-parametrized daq2 testing for now. I'll add a non-complex equivalent for the iq loopback test that will be used in the daq2.

By the way, I had to modify the dds.py single tone and dual tone generation methods. I had to add a way to access the chip_b dds similar to how its done in the __update_dds but not in a loop.

Thanks.

PR #66, opened 2020-02-20
MxFE

mhennerich:


tfcollins:
I need to verify some changes to move away from the hard coded channels before merging this. Some of the get attribute calls return too many duplicate results. I will verify with the hardware in GBO tomorrow.

mhennerich:
Sounds good! 
Thanks

PR #70, opened 2020-02-27
Added AD9152 and DAQ3

kister-jimenez:
- Added AD9152 and Daq3 along with their parametrized tests
- Added AD9364/ADRV9364 tests.
- Added buffer sizes setting in DMA and DDS rx test routines.
- Changed the iq test routine so it can handle non-complex device DMA testing. Renamed it to cw_loopback instead of iq_loopback. 
- Remove checking if number of channels is greater than 2 before enabling channels in conftest. This causes non-complex devices with 2 channels to get all channels enabled (even if only 1 channel should be enabled).


tfcollins:
Can some of these commits be squashed together?

kister-jimenez:
I tried squashing then I did a rebase up until the previous PR to clean my commits. It's my first time to do this so I hope that what I'm doing follows the correct flow.

tfcollins:
Much better :). Just two things, can you update commit messages 2 and 3 to include a message about the addition of the part interfaces.You did more work than just add tests.

PR #72, opened 2020-03-05
Add support for LTC2983

airwoodix:
This patch adds support for the multi-sensor temperature measurement IC [LTC2983](https://www.analog.com/en/products/ltc2983.html).

Heavily inspired by the AD7124 interface. AFAICT, there's no buffer support for LTC2983 but I nevertheless kept support for it in the interface (`rx` subclass).

Also includes very minimal examples, very hardware dependent.

Thanks for your time!

### DIFF
--- a/adi/ltc2983.py
+++ b/adi/ltc2983.py
@@ -1,9 +1,8 @@
+import numbers
 from collections import OrderedDict
 from collections.abc import Iterable
-import numbers
 
 import numpy as np
-
 from adi.attribute import attribute
 from adi.context_manager import context_manager
 from adi.rx_tx import rx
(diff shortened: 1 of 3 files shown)


adi/ltc2983.py
```diff
@@ -0,0 +1,79 @@
...
+from adi.attribute import attribute
+from adi.context_manager import context_manager
+from adi.rx_tx import rx
+
+
+class ltc2983(rx, context_manager):
+    """ LTC2983 Multi-Sensor Temperature Measurement System """
+
+    _complex_data = False
+    channel = None
```
tfcollins: This needs an annotation `channel: OrderedDict = None`

adi/ltc2983.py
```diff
@@ -0,0 +1,79 @@
...
+        def __init__(self, ctrl, channel):
+            self._ctrl = ctrl
+            self._channel = channel
+
+            self.name = channel.id
+
+            # raw value attribute is '<x>_raw' with
+            # <x>=thermistor,rtd,diode,thermocouple,direct_adc
+            raw_attr_name = [x for x in channel.attrs.keys() if x.endswith("raw")]
+            assert len(raw_attr_name) == 1
```
tfcollins: Can you add an error message to the assert

airwoodix: Removed this. I was working against an older revision of the kernel driver. The newest version analogdevicesinc/linux@2b453a1 fixes this with a simpler attribute naming scheme.

PR #73, opened 2020-03-06
Adding ADRV9361-z7035 RF manufacturing tests.

fhurgoi:
Changes to ADRV9009_ZU11EG RF manufacturing tests for tesing Observation receive channels.


test/rf/spec.py
```diff
@@ -126,10 +126,10 @@ def find_harmonics(x, freqs, num_harmonics=6, tolerance=0.01):
     return main, main_loc, harmonics_vals, harmonics_locs
 
 
-def sfdr(x, fs=1, ref=2 ** 15, plot=False):
+def sfdr(x, fs=1, ref=2 ** 15, plot=True):
```
tfcollins: Plots should remain off by default. This can break headless builds

test/conftest.py
```diff
@@ -328,37 +328,38 @@ def t_sfdr(classname, devicename, channel, param_set, sfdr_min):
...
+    N = 2 ** 15
     ts = 1 / float(RXFS)
     t = np.arange(0, N * ts, ts)
-    i = np.cos(2 * np.pi * t * fc) * 2 ** 15 * 0.9
-    q = np.sin(2 * np.pi * t * fc) * 2 ** 15 * 0.9
+    i = np.cos(2 * np.pi * t * fc) * 2 ** 15 * 0.5
+    q = np.sin(2 * np.pi * t * fc) * 2 ** 15 * 0.5
     iq = i + 1j * q
     # Pass through SDR
+    window = np.hanning(sdr.rx_buffer_size)
```
tfcollins: Can we move this into the spectral estimation function?
For example enable it here: https://github.com/analogdevicesinc/pyadi-iio/blob/master/test/rf/spec.py#L26

test/conftest.py
```diff
@@ -328,37 +328,38 @@ def t_sfdr(classname, devicename, channel, param_set, sfdr_min):
     # Set custom device parameters
     for p in param_set.keys():
         setattr(sdr, p, param_set[p])
-    time.sleep(5)  # Wait for QEC to kick in
+    # time.sleep(5)  # Wait for QEC to kick in
```
tfcollins: This breaks ADRV9009 tests. QEC takes forever on those parts

test/conftest.py
```diff
@@ -328,37 +328,38 @@ def t_sfdr(classname, devicename, channel, param_set, sfdr_min):
...
     sdr.tx_cyclic_buffer = True
-    sdr.tx_enabled_channels = [channel]
-    sdr.tx_buffer_size = N * 2 * len(sdr.tx_enabled_channels)
-    sdr.rx_enabled_channels = [channel]
-    sdr.rx_buffer_size = N * 2 * len(sdr.rx_enabled_channels)
+    sdr.rx_buffer_size = 2 ** 15
+    if sdr._num_tx_channels > 2:
+        sdr.tx_enabled_channels = [channel]
+    if sdr._num_rx_channels > 2:
+        sdr.rx_enabled_channels = [channel]
```
kister-jimenez: Did the previous line 333-338 break your code? I tried to remove the "if sdr._num_tx_channels > 2:" so that I can enable 1 channel only from a none complex-data device that only have 2 TX channels like DAQ2.

tfcollins: So we don't have any SFDR tests for non-complex devices yet, but in that case we might want to implement a single sided SFDR measurement. Then you don't have to use DAC/ADC pairs.

adi/adrv9009_zu11eg.py
```diff
@@ -193,3 +193,72 @@ def trx_lo_chip_b(self):
...
+        """Powerdown:  Eanbles/Disables the Observation Receive Chain"""
+        return self._get_iio_attr("voltage3", "powerdown", False, self._ctrl_b)
+
+    @obs_powerdown_chan1_ctrl_b.setter
+    def obs_powerdown_chan1_ctrl_b(self, value):
+        """Powerdown:  Eanbles/Disables the Observation Receive Chain"""
+        return self._set_iio_attr("voltage3", "powerdown", False, value, self._ctrl_b)
+
+    @property
+    def obs_hardwaregain_chan0_ctrl_b(self):
```
tfcollins, line 249: Need to be  obs_hardwaregain_chan0_chip_b not obs_hardwaregain_chan0_ctrl_b. Same for others below

PR #82, opened 2020-05-05
Add test configuration and related doc

tfcollins:
PR adds initial test configuration from yaml which is very handy for CI and generally development. Included is also doc for developers who would use these config files, along with other details for developers on tools.

### DIFF
--- a/test/iio_scanner.py
+++ b/test/iio_scanner.py
@@ -199,7 +199,7 @@ def get_device(uri):
     return None
 
 
-def find_device(names, uri=None, config=None):
+def find_device(names, uri=None, config=None, ignore_skip=None):
     skip_usb = False
     if not isinstance(names, list):
         names = [names]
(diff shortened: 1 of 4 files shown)


test/iio_scanner.py
```diff
@@ -209,6 +209,15 @@ def find_device(names, uri=None):
...
         else:
             return (False, [])
+    elif config:
+        # Convert config to board class
+        boards = []
+        for dev in config["uri-map"]:
+            hardware_options = config["uri-map"][dev]
+            for hardware in hardware_options.split(","):
+                b = board(hardware, dev)
+                boards.append(b)
```
kister-jimenez, line 219: This will cause the test to fail if the hardware and uri combination is not connected or online even without the --error_on_filter option

tfcollins: This should be fixed now

PR #99, opened 2020-06-11
Added __repr__ to the Pluto object with some key parameters.

petrushy:
Hi, I think adding some human readable __repr__ strings to key objects is good for usage in interactive environments like jupyter.

Here is a first attempt for the Pluto object. I think some more parameters could be useful but then it is outside the pyadi-iio abstractions.

This one needs python 3.6 at least, don't know what your compatibility targets are.

### DIFF
--- a/adi/ad936x.py
+++ b/adi/ad936x.py
@@ -288,3 +288,23 @@ class Pluto(ad9364):
 
     _device_name = "PlutoSDR"
     _uri_auto = "ip:pluto.local"
+
+    def __repr__(self):
+        retstr = f"""Pluto() object "{self._device_name}" with following key properties:
+
+rx_lo:                   {self.rx_lo / 1000000:<12} (MHz, Carrier frequency of RX path)
+rx_hardwaregain:         {self.rx_hardwaregain:<12} (dB, Gain applied to RX path. Only applicable when gain_control_mode is set to 'manual')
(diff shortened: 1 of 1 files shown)


adi/ad936x.py
```diff
@@ -288,3 +288,23 @@ class Pluto(ad9364):
 
     _device_name = "PlutoSDR"
     _uri_auto = "ip:pluto.local"
+
+    def __repr__(self):
+        retstr = f"""Pluto() object "{self._device_name}" with following key properties:
+
+rx_lo:                   {self.rx_lo / 1000000:<12} (MHz, Carrier frequency of RX path)
+rx_hardwaregain:         {self.rx_hardwaregain:<12} (dB, Gain applied to RX path. Only applicable when gain_control_mode is set to 'manual')
```
tfcollins, line 296: Some of these attributes don't exist in the current classes and instead have an appended _chan0 to them. To get this to work please update these.

petrushy: Hi, wasn't sure about this _chan0, saw some (old?) documentation not using it and seemed to work. As it is abstraction for a specific device (Pluto) that only have one (?) channel I thought it made some sense to abstract them away. Have added them now back.

petrushy:
Sorry, messed up with git and signing. Will make new PR.

PR #100, opened 2020-06-15
Extended __repr__ of Pluto object

petrushy:
An improved __repr__ for the adi.Pluto() object.

Now gpg signed.

### DIFF
--- a/adi/ad936x.py
+++ b/adi/ad936x.py
@@ -288,3 +288,23 @@ class Pluto(ad9364):
 
     _device_name = "PlutoSDR"
     _uri_auto = "ip:pluto.local"
+
+    def __repr__(self):
+        retstr = f"""Pluto(uri="{self.uri}") object "{self._device_name}" with following key properties:
+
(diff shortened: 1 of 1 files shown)


tfcollins:
Can you update the commit message to include: Signed off by **
Look at the commit here as an example: https://github.com/analogdevicesinc/pyadi-iio/commit/09dd3ebb91c39e652a488b450e0202629be4727d

-Travis

rgetz:
There is a difference between "signing" your commit with pgp/gpg ([Doc](https://help.github.com/en/github/authenticating-to-github/signing-commits)) and adding a `"Signed-off-by"` 

```
-s
--signoff
Add Signed-off-by line by the committer at the end of the commit log message.
```
which means you agree with [Developer's Certificate of Origin 1.1](https://developercertificate.org/)

It's the latter we care about.

petrushy:
Should be properly signed now I think.

rgetz:
It is - thanks alot.

PR #102, opened 2020-06-26
adis16507: Remove burst_mode_enable

nunojsa:
This attribute was removed from the driver. Burst mode is just used by
default.

Signed-off-by: Nuno Sá <nuno.sa@analog.com>


tfcollins:
Thanks. I updated the example that cause CI to fail

PR #129, opened 2020-10-22
Speed up, fix, and clarify adar1000 class

wsapia-adi:
# Description
- Fixes an issue with the number of controls found vs. number of controls requested
- Clarifies input/output state for the _add_prop method
- Reverses polarity of the _add_prop method since before it was marking outputs as "rx"
- Speeds up a couple of loops on instantiation

## Type of change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How has this been tested?
I tested this by instantiating 8 separate beams at once.

**Test Configuration**:
* Hardware: M2K
* OS: Windows 10

# Checklist:


adi/adar1000.py
```diff
@@ -55,17 +55,17 @@ def __set__(self, instance, value):
         )
 
 
-def _add_prop(classname, channelname, attr, i, inout, dev, beam_name=None):
+def _add_prop(classname, channelname, attr, i, output, dev, beam_name=None):
```
wsapia-adi, line 58: changed "inout" since it's not clear whether inputs should be True or False

adi/adar1000.py
```diff
@@ -100,7 +100,8 @@ def __init__(self, uri="", beams="BEAM0"):
                 for dev in self._ctx.devices:
                     if "label" in dev.attrs and dev.attrs["label"].value == beam:
                         self._ctrls.append(dev)
-            if len(self._ctrls) == len(beams):
+                        break
```
wsapia-adi, line 103: The loop doesn't need to keep checking devices if we've already found the right one.

adi/adar1000.py
```diff
@@ -100,7 +100,8 @@ def __init__(self, uri="", beams="BEAM0"):
...
+            if len(self._ctrls) != len(beams):
```
wsapia-adi, line 104: If the lengths are equal, then all the beams were found

adi/adar1000.py
```diff
@@ -114,8 +115,8 @@ def __init__(self, uri="", beams="BEAM0"):
             beams = [None]
             self._ctrls = [self._ctrl]
         for b, beam in enumerate(beams):
+            dev = self._ctrls[b]
```
wsapia-adi, line 118: No need to keep defining this in the inner loop

wsapia-adi:
@tfcollins 

I'm not sure exactly how to fix this issue. Any chance you could take a look?

adi/adar1000.py
```diff
@@ -125,60 +128,78 @@ def __init__(self, uri="", beams="BEAM0"):
...
+            )
+            for dev in self._ctrls
+            for b in self._beam_channels
+        }
+
+    def _set_attribute_dictionary(self, attribute_name, output, values):
+        """ Set all requested attributes by name and output style using a dictionary """
+
+        # If all the new values are the same as the first value, set them all at once
+        if all([next(iter(values.values())) == v for v in values.values()]):
```
tfcollins, line 183: This check really just makes the code complex. Under the hood, the module will just break stuff down and do sequential calls so there is no speed advantage. Just having the loop below will get to the same result

adi/adar1000.py
```diff
@@ -125,60 +128,78 @@ def __init__(self, uri="", beams="BEAM0"):
...
+        # If there are different values, set them one at a time.
+        else:
+            for key, value in values.items():
+                # Pull out the beam and channel names
+                label_sections = key.lower().split("_")
+                beam = "_".join(label_sections[:2])
+                channel = label_sections[-1]
+
+                # Determine the device
+                dev = [d.attr["label"].value.lower() == beam for d in self._ctrls][0]
```
tfcollins, line 201: 1. d.attr -> d.attrs
2. This returns a boolean. It needs to return a iio_device object
3. Needs a check to make sure dev is not null (otherwise set_iio_attr will default to self._ctrl)

adi/adar1000.py
wsapia-adi: That's fine with me, I just left it because the original code worked that way. I'll remove it.

adi/adar1000.py
wsapia-adi: Yep, good catch. I'll fix this now.

adi/adar1000.py
```diff
@@ -125,60 +128,65 @@ def __init__(self, uri="", beams="BEAM0"):
...
+            beam = '_'.join(label_sections[:2])
+            channel = label_sections[-1]
+
+            # Determine the device & set the attribute
+            try:
+                dev = [c for c in self._ctrls if c.attrs['label'].value.lower() == beam][0]
+                self._set_iio_attr(channel, attribute_name, output, value, dev)
+            except IndexError:
+                pass
+        return
```
tfcollins, line 192: Don't need the return.

adi/adar1000.py
```diff
@@ -125,60 +128,65 @@ def __init__(self, uri="", beams="BEAM0"):
...
+            # Pull out the beam and channel names
+            label_sections = key.lower().split('_')
+            beam = '_'.join(label_sections[:2])
+            channel = label_sections[-1]
+
+            # Determine the device & set the attribute
+            try:
+                dev = [c for c in self._ctrls if c.attrs['label'].value.lower() == beam][0]
+                self._set_iio_attr(channel, attribute_name, output, value, dev)
+            except IndexError:
```
tfcollins, line 190: Can you raise a custom exception here. Otherwise, it will fail without notification

wsapia-adi: will do

PR #130, opened 2020-10-27
Multi som adrv9009 jesd204 fsm

mhennerich:
# Description

With the new jesd204-fsm multichip_sync writes are no longer required.
Clock chip synchronization is done inside the framework.
Synchronizing distributed multi-topology systems, requires some extra
functionality in jesd204_fsm_sync() which resumes a number of
iio devices across different IIO context from the jesd204-fsm stop-states.

This code is intended to sync multiple FPGA systems across
jesd204-fsm topologies.

There is always one primary device, the device which controls the
sysref-provider and an open list of secondary devices which are synced
by the same sysref-provider connected clock source.

Fixes # (issue)

## Type of change

# How has this been tested?

ADRV9009-ZU11EG System on Module on the ADRV2CRR-FMC carrier board with an additional FMCOMMS8 FMC board connected. So, in total there are 4x ADRV9009 RF transceivers in this design, with a two-level clock tree.

# Documentation

https://wiki.analog.com/resources/tools-software/linux-drivers/jesd204/jesd204-fsm-framework#synchronizing_distributed_multi-topology_systems

# Checklist:


examples/adrv9009_som_multi.py
```diff
@@ -0,0 +1,169 @@
...
+
+multi._dma_show_arming = False
+multi._jesd_show_status = True
+multi._jesd_fsm_show_status = True
+multi._clk_chip_show_cap_bank_sel = True
+multi._resync_tx = True
+multi.rx_buffer_size = 2 ** 10
+
+multi.primary._clock_chip_ext.reg_write(0xCF, 1)
+multi.primary._clock_chip_ext.reg_write(0xE3, 1)
```
tfcollins, line 71: Can we add a method for this?

mhennerich: Totally agree - ideally we have a hmc7044 class.
But having a method with takes the output channel as parameter and a delay setting is definitely good to have.

tfcollins: Yes, a class for 7044 would be nice. Would be want to put these functions in the driver first though? I don't know if we would want a class that just has a bunch of register writes in it.

mhennerich: > Would be want to put these functions in the driver first though? 

Well they are already. But the digital delay is expressed as phase in RAD and changes by frequency. Not something that you want if you just want to specify a delay.
Both digital and analog delay can be set via devicetree for each channel.
The function here is to simply determine values which work and can be put into the devicetree afterwards.

tfcollins:
@mhennerich can you double-check I didn't add a breaking change on your setup. I make the JESD monitors optional in the core classes which speeds up init a lot.

mhennerich:
> @mhennerich can you double-check I didn't add a breaking change on your setup. I make the JESD monitors optional in the core classes which speeds up init a lot.

Seems to work ok

PR #139, opened 2020-11-13
Update ADAR1000 wrapper & add GPIO wrapper

wsapia-adi:
# Description

Changed ADAR1000 driver to actively get all the device and channel attributes.

Still working with SDG to correctly implement all required attributes.

## Type of change

# How has this been tested?

**Test Configuration**:
* Hardware: Analog Discovery Kit
* OS: Windows 10

# Documentation

- Removed access to all hardwaregains and phases as that wasn't realistically how people should be accessing those attributes.
- Wrapper now dynamically adds all attributes with descriptive names

# Checklist:


tfcollins:
A few things here:
1. AD9081 is not ready to be merged. There are a number of features missing from the driver and the initialization will fail in different DDC/UDC configurations. This is being addressed with help of a driver update. That class needs to be removed.
2. We try to avoid having fully dynamic properties for classes since it removes any doc related to those properties. There are also use cases where you want to write to all the phases and gains in a single call. This is common for MIMO systems where gains are calculated in a vectorized way. Having to loop through the attributes to set them all is rather painful.
3. Class additions should be split into individual commits. This makes cherry-picking easier
4. Testing against `Analog Discovery Kit` requires none of this code... Was this run against an actual board?

adi/one_bit_adc_dac.py
```diff
@@ -0,0 +1,103 @@
...
+            Dynamic class properties will be created for each channel.
+    """
+
+    _device_name = ""
+
+    def __init__(self, uri="", name=""):
+
+        # If no name is given, this will choose the first device of the correct type
+        if name == "":
+            name = "one-bit-adc-dac"
```
tfcollins, line 75: Set default value of name in __init__ function definition to "one-bit-adc-dac". Has same behavior with 3 less lines.

adi/adar1000.py
```diff
@@ -50,150 +69,121 @@ def __get__(self, instance, owner):
...
+            four ADAR1000s. Use a list when multiple chips are to be
+            controlled. Dynamic class properties will be created for each
+            ADAR1000 device.
     """
 
     _device_name = ""
-    _beam_channels = ["voltage0", "voltage1", "voltage2", "voltage3"]
 
-    def __init__(self, uri="", beams="BEAM0"):
+    def __init__(self, uri="", chip_ids="csb1_chip1"):
```
tfcollins, line 106: Does this value map to the default driver name for ADAR1000? or is the specific to stingray?

adi/one_bit_adc_dac.py
```diff
@@ -0,0 +1,103 @@
...
+            self._channel_name, self._attr, self._output, int(value), self._dev
+        )
+
+
+class one_bit_adc_dac(attribute, context_manager):
+    """One bit ADC/DAC (GPIO)
+
+    parameters:
+        uri: type=string
+            URI of IIO context with ADAR100(s)
```
tfcollins, line 63: ADAR1000(s) -> One-bit-adc-dac(s)

wsapia-adi:
Hey Travis,
1. I misunderstood our earlier chat, I thought you said that if I included the AD9081, it would be merged. That's fine, I'll remove it.
2. I thought the combined gain/phase property was more confusing than helpful. If someone wanted this capability, couldn't they just add it to their own wrapper which inherits this one? I figured that it's better to make the class more ADAR1000-specific than array-specific and let the user build up from there if they want to.
3. I'll try to do this in the future.
4. I'm still testing the IIO driver directly and updating Mircea with any required changes. There's still some work left to do on that front. This pull request essentially just reconfigures the way the driver presents the attributes, but I've verified that the commands work as you'd expect, even if the underlying IIO driver isn't always doing the right thing.

adi/one_bit_adc_dac.py
wsapia-adi: You're right, brain-fart. I'll fix this.

adi/adar1000.py
wsapia-adi: I'm going to recommend that the default is updated to this style since it's more explicit and easier to understand at a glance. The "BEAM" notation doesn't tell you anything about how to access the chip whereas this notation tells you everything about how to interface with that particular IC.

adi/one_bit_adc_dac.py
wsapia-adi: Yep, copy/paste error. I'll fix it.

wsapia-adi:
As for the dynamic properties. I'm happy to make them all static, but it's going to make the driver very verbose.

Where would you expect to see this documentation? Comments in the code, or explicitly somewhere else? I'm all for more documentation, and I'm even happy to write it, I just want to make sure it's done in the proper place. The original ADAR1000 documentation had very little information, so there isn't much to go on there.

tfcollins:
You can see the doc here where the gain and phases doc show up: https://analogdevicesinc.github.io/pyadi-iio/devices/adi.adar1000.html
This gets generated from the doc strings.
This is what we always try to reference in support questions.

Can you update the PR text to mention you are going back and forth with Mircea.

wsapia-adi:
If you want explicit documentation for each parameter, this file is going to be quite verbose. As long as that's okay with you, I'm happy to refactor the code to do this. It will take some time.

wsapia-adi:
Statically adding properties will likely require individual handles for each ADAR1000 IC rather than accessing all devices through one handle. That will be quite slow to instantiate for large arrays. Should we consider two wrappers? Is there another option for documentation that we can implement to overcome this?

adi/adar1000.py
```diff
@@ -35,165 +35,858 @@
...
-    def rx_phases(self, values):
-        """rx_phases: Set all phases of RX path"""
-        self._set_attribute_dictionary("phase", False, values)
-
-    def _get_attribute_dictionary(self, attribute_name, output):
-        """ Get a dictionary of all attached requested attributes by name and output style """
+            output = True
+
+        self._set_iio_attr("voltage0", "beam_pos_load", output, state, self._ctrl)
+        return
```
tfcollins, line 673: returns are not needed in these methods

adi/__init__.py
```diff
@@ -73,9 +73,11 @@
 
 from adi.ad5686 import *
 
-from adi.adar1000 import adar1000
+from adi.adar1000 import *
```
tfcollins, line 76: Please make the imports explicit

PR #180, opened 2021-02-25
fix/add logos

rgetz:
# Description

adds some svgs and pngs of logos with white/grey/black text.


rgetz:
update with newlines at the end of the svg (which isn't necessary, and isn't added by inkscape), so that the style checking is happy.

PR #182, opened 2021-03-02
Added pyadi-iio support files for AD7606 devices

mphalke:
Signed-off-by: mahphalke <Mahesh.Phalke@analog.com>

# Description
1) Added supported device and channel attributes through adi/ad7606.py file.
Reference: https://wiki.analog.com/resources/tools-software/linux-drivers/iio-adc/ad7606
2) Updated other supported files (example/ad7606.py, .rst, etc).

Fixes # (issue)

## Type of change

# How has this been tested?
The code was tested with ARM based firmware application which uses Tinyiiod libraries.
Reference: https://wiki.analog.com/resources/tools-software/product-support-software/ad7606_mbed_iio_application

**Test Configuration**:
* Hardware: EVAL-AD7606B-FMCZ
* OS: Windows

# Documentation

# Checklist:

- [ ] I have signed off all commits and they contain "Signed-off by: <Mahesh.Phalke@analog.com>"

### DIFF
--- a/adi/ad7606.py
+++ b/adi/ad7606.py
@@ -74,17 +74,6 @@ class ad7606(rx, context_manager):
 
         rx.__init__(self)
 
-    def rx(self):
-        sig = super().rx()
-
-        if (
-            self._rx_unbuffered_data
-            or self._complex_data
-            or self.rx_output_type == "raw"
-        ):
-            return sig
-            
-        return [signal/1000 for signal in sig]
 
     @property
     def scale_available(self):
--- a/adi/ad7606.py
+++ b/adi/ad7606.py
@@ -83,13 +83,8 @@ class ad7606(rx, context_manager):
             or self.rx_output_type == "raw"
         ):
             return sig
-        else:
-            mv_sig = []
-
-            for signal in sig:
-                mv_sig.append(signal / 1000)
-
-            return mv_sig
+            
+        return [signal/1000 for signal in sig]
 
     @property
     def scale_available(self):
--- a/adi/ad7606.py
+++ b/adi/ad7606.py
@@ -1,0 +1,160 @@
+# Copyright (C) 2021 Analog Devices, Inc.
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification,
+# are permitted provided that the following conditions are met:
+#     - Redistributions of source code must retain the above copyright
+#       notice, this list of conditions and the following disclaimer.
+#     - Redistributions in binary form must reproduce the above copyright
+#       notice, this list of conditions and the following disclaimer in
+#       the documentation and/or other materials provided with the
+#       distribution.
+#     - Neither the name of Analog Devices, Inc. nor the names of its
+#       contributors may be used to endorse or promote products derived
+#       from this software without specific prior written permission.
+#     - The use of this software may or may not infringe the patent rights
+#       of one or more patent holders.  This license does not release you
+#       from the requirement that you obtain separate licenses from these
+#       patent holders to use this software.
+#     - Use of the software either in source or binary form, must be run
+#       on or directly connected to an Analog Devices Inc. component.
+#
+# THIS SOFTWARE IS PROVIDED BY ANALOG DEVICES "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES,
+# INCLUDING, BUT NOT LIMITED TO, NON-INFRINGEMENT, MERCHANTABILITY AND FITNESS FOR A
+# PARTICULAR PURPOSE ARE DISCLAIMED.
+#
+# IN NO EVENT SHALL ANALOG DEVICES BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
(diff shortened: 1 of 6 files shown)


tfcollins (changes requested):
Thanks a lot for the addition :)

Can you add some basic testing for the part in the tests folder?  Just configuring  test_dma_rx and  test_attribute_single_value should be enough.

There are also some linter errors (run `pre-commit run --all-files`)

```
codespell................................................................Passed
isort....................................................................Failed
- hook id: isort
- files were modified by this hook

Fixing /tmp/pyadi-iio-pr182/adi/ad7606.py
Skipped 2 files
Skipped 1 files

Check Yaml...............................................................Passed
Fix End of Files.........................................................Passed
Trim Trailing Whitespace.................................................Passed
Check for merge conflicts................................................Passed
Debug Statements (Python)................................................Passed
Check docstring is first.................................................Passed
Flake8...................................................................Passed
black....................................................................Failed
- hook id: black
- files were modified by this hook

reformatted /tmp/pyadi-iio-pr182/adi/ad7606.py
All done! ✨ 🟟 ✨
1 file reformatted, 102 files left unchanged.

blacken-docs.............................................................Passed
```

adi/ad7606.py
```diff
@@ -0,0 +1,160 @@
...
+    def rx(self):
+        sig = super().rx()
+
+        if (
+            self._rx_unbuffered_data
+            or self._complex_data
+            or self.rx_output_type == "raw"
+        ):
+            return sig
+        else:
```
tfcollins, line 86: The else is unnecessary here. You would do this in a little list comprehension too like: 
```python
return [sig/1000 for sig in signal]
```

mphalke: I referred this code from ad7124.py module. Below is my justification for this part of code. Please correct me if I am wrong.
The 'if' part is used to return ADC read data just in raw format provided 'rx_output_type' type is "raw" Or '_rx_unbuffered_data' is true or '_complex_data' is true. I think '_complex_data' might be redundant here as it is local to ad7606 class and set to 'false' during init. May be this can be removed from 'if' check?
The other two attributes are defined in 'rx_tx.py'. So if someone want's read data in voltage converted form, they can set '_rx_unbuffered_data = true' Or 'rx_output_type = "SI"'. Again this is referred from ad7124.py. The else part would then convert raw data into eqv voltage (by considering scale factor).
Let me know your thoughts. I have added test results below (this is for my ARM based firmware example and not on FPGA):

image

mphalke:
Hi @tfcollins 
I have added test_ad7606.py file as per your suggestions. Though I am not sure what different attributes I can add for tests. Seems like only device attribute for ad7606b that can be modified is 'oversampling_ratio' (as listed in linux  drivers).
I did not understand what 'step' value should I put for OSR as OSR value can be any of 1,2,4,8,16,32,64,256, so step size is 2^x, where x is OSR. For simple testing I have added start=1, stop=2 and step=1. Does that make sense?

tfcollins:
For powers of two you can use attribute_single_value_pow2

-Travis

adi/ad7606.py
tfcollins: Sorry my comment wasn't clear. The separate conditions are fine but since there is a return you don't need to use an else for the second condition. It should be something like:

if (self._rx_unbuffered_data
            or self._complex_data
            or self.rx_output_type == "raw"
        ):
     return sig

return [sig/1000 for sig in signal]

mphalke: Okay got it now, sorry for the confusion. I will make this change.
Scale attribute is not set to 1/10000. So I think that's why result needs to be divided by 1000.

https://wiki.analog.com/resources/tools-software/linux-drivers/iio-adc/ad7606
image

tfcollins: Sorry for the delay. I was out on vacation.

So after applying the scale and bias it should result in a millivolt. Is that not the case?

mphalke: By looking at AD7606 Linux drivers, the scale is represented to obtain the results in millivolt. 
For example: scale = range / adc resolution = 10v / 2^16 = 10/65535 = 0.000152
However, range factor is multiplied by 1000 in Linux drivers, so the scale becomes = 10000/65535 = 0.152.
And then formula used is: U = in0_raw * in_scale = 5789 * 0.152 = 879.928 mV

So, your comment above is valid i.e. after multiplying scale with adc raw value, the result should be in millivolts. And if that is the case, the additional code used to convert raw value into millivolt (i.e. signal/1000) is unnecessary.

The function would then look like as below, right?

**def rx(self):
        return super().rx()**

I would wait for your inputs on this before committing code, just to make sure my understanding is correct.

tfcollins: kk. So if that is the case you don't even need the rx method in your class since it will just inherit it.

-Travis

mphalke: Agree with you, this method won't require. Do you want me to delete this method completely from ad7606.py class file?

tfcollins: Yes. That should be the last thing and I'll merge. Thanks for the contribution :+1: 

-Travis

PR #204, opened 2021-06-23
Add ADF4371 compatibility

wsapia-adi:


mhennerich (changes requested):
I think this was fixed in the driver with this commit:
https://github.com/analogdevicesinc/linux/commit/dc6dabe412bf60dbc52a8c9dba64ee1d272bbe73
So - no need to defeature temperature...

wsapia-adi:
I didn't see temperature in the attributes when I was poking at the driver, so I removed it. I also had a copy/paste error in the code, so it was incorrect anyway. I can add it back assuming the attribute is named "in_temp0_raw".

adi/adf4371.py
```diff
@@ -174,3 +174,8 @@ def rf32_frequency(self):
     def rf32_frequency(self, value):
         """Get/Set the frequency of the 32GHz RF output"""
         self._set_iio_attr("altvoltage3", "frequency", True, value, self._ctrl)
+
+    @property
+    def temperature(self):
+        """Get the temperature reading"""
+        return self._get_iio_attr("temp0", "raw", False)
```
tfcollins, line 181: @mhennerich is this the correct dev+attr name?

mhennerich: The driver uses IIO_CHAN_INFO_PROCESSED
https://github.com/analogdevicesinc/linux/blob/master/drivers/iio/frequency/adf4371.c#L764

So the full channel name is: **in_temp0_input**

So this should be 
return self._get_iio_attr("temp0", "input", False)

tfcollins:
Moved PR here #210  with updates

PR #228, opened 2021-09-14
Update save to eeprom functionality

ranechita:
Signed-off-by: Ramona Alexandra Nechita <ramona.nechita@analog.com>

# Description

When running the production tests, at the end of the calibration procedure a window is prompted to introduce the serial number of the device and then eeprom is written accordingly. All functions related to the eeprom are moved to a separate file. It also includes minor fixes.

## Type of change

**Test Configuration**:
* Hardware: ADRV9361-ZC702
* OS: Ubuntu 20.04

# Documentation

# Checklist:

### DIFF
--- a/test/eeprom.py
+++ b/test/eeprom.py
@@ -1,0 +1,117 @@
+import tkinter as tk
+from datetime import date, datetime
+from tkinter import Button, Entry, Label
+
+import paramiko
+import pytest
+
+
+def save_to_eeprom(iio_uri, coarse, fine):
+    full_uri = iio_uri.split(":", 2)
+    if full_uri[0] != "ip":
+        pytest.skip("Tuning currently supported only for ip URIs")
+    ip = full_uri[1]
+
+    ssh_client = paramiko.SSHClient()
+    ssh_client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+    ssh_client.connect(ip, username="root", password="analog")
+    ssh_stdin, ssh_stdout, ssh_stderr = ssh_client.exec_command(
+        "find /sys/ -name eeprom"
+    )
+    eeprom_path = ssh_stdout.readline().rstrip("\n")
+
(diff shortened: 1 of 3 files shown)


test/eeprom.py
```diff
@@ -0,0 +1,114 @@
...
+    if ssh_stderr.channel.recv_exit_status() != 0:
+        raise paramiko.SSHException("find_eeprom command did not execute properly")
+
+    this_day = date.today()
+    if this_day.year <= 1995:
+        pytest.skip(
+            "System date and time not in order. PLease connect the device to internet to update."
+        )
+
+    sn = popup_txt()
```
tfcollins, line 32: Can we add a check here to make sure the input is reasonable

test/scpi.py
```diff
@@ -185,7 +116,6 @@ def dcxo_calibrate(classname, iio_uri):
 
     sdr._set_iio_dev_attr("dcxo_tune_coarse", coarse, sdr._ctrl)
     sdr._set_iio_dev_attr("dcxo_tune_fine", fine, sdr._ctrl)
-
     while 1:
         current_frq = get_freq(instr)
```
tfcollins, line 120: Any scenario when this would never exit?

test/eeprom.py
```diff
@@ -0,0 +1,114 @@
+import tkinter as tk
+from datetime import date, datetime
+from tkinter import Button, Entry, Label
```
tfcollins, line 3: Will need a separate requirements.txt file (maybe requirements_prod_test.txt). TK doesn't always ship with python

test/eeprom.py
ranechita: I added a check in the popup_txt() function, right before returning the value

test/scpi.py
ranechita: I don't think there is any

test/eeprom.py
ranechita: I created a new commit to add that

PR #235, opened 2021-09-21
adrv9002 sdk 0.17

nunojsa:
# Description

This updates the adrv9002 class tests to be compatible with the SDK integrated in the driver. The main difference is related with digital interface gain. It is now clear that the RX gain table types cannot change at runtime so that the ABI for the digital gain changed and that has to be updated in the tests. Furthermore, the compensated gain tables are still not supported so the related tests were removed.

The second major thing is the refactor of the DMA tests. This is actually a bugfix as the tests were failing for split mode (when channel > 0) and we were not testing RX2/TX2 in MIMO mode. The change unifies the DMA tests so that we do not need different tests and HW markers for the different DMA modes (more info in the commit message).

## Type of change

# How has this been tested?

Tested running the unit tests defined for the adrv9002 class.

**Test Configuration**:
* Hardware: ZynMP 
* OS: ADI Kuiper

# Checklist:


nunojsa:
> Are the old AGC modes completely gone now?

The ABI as it was defined it's gone (it was wrong)... The problem was that the digital gain API had a parameter so that you could change the table type between correction and compensated: Thus, the four possibilities:

1. Gain_Correction_manual_control;
2. Gain_Compensation_manual_control;
3. Gain_Compensation_automatic_control;
4. Gain_Correction_automatic_control.

Having this on the API, led me to assume that we could change this at runtime... It turns out we cant and that is made clear with the new API were the parameter to change the gain tables is gone. The compensated vs correction tables types has to be decided at initialization (when loading the profile) but as I mentioned, we are still not getting that info from the profile so that we are just loading the default table which is correction (that's why some tests were just removed).

Anyways, the end result is that you can only change between manual and automatic control for the digital gain (the same as for the primary gain control)... The table type is decided at profile load. When things are properly supported, I might export this information when reading the profile.

> Do we need to remove checks like: https://github.com/analogdevicesinc/pyadi-iio/blob/master/adi/adrv9002.py#L301 ?

Oh definitely. And this actually explains an odd behavior I was seeing last night. I was just preparing to debug this but now you gave me the answer :)

nunojsa:
v2:

* Fixed adrv9002 class with respect with digital gain control...

PR #236, opened 2021-09-30
adi/QuadMxFE_multi: Add Multi QuadMxFE support

mhennerich:
# Description

 * Add JESD204-FSM controls to ad9081 base class
 * Add HMC7043 clock chip to QuadMxFE class
 * Add QuadMxFE_multi class
 * Add Quad_MxFE_dual_example using the QuadMxFE_multi class

## Type of change

# How has this been tested?

Tested on HW
# Checklist:

### DIFF
--- a/adi/QuadMxFE_multi.py
+++ b/adi/QuadMxFE_multi.py
@@ -1,0 +1,366 @@
+# Copyright (C) 2021 Analog Devices, Inc.
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification,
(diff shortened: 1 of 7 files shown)


tfcollins:
LGTM. This brings to light that we probably need to create a JESD FSM class to handle that more generally.

mhennerich:
> LGTM. This brings to light that we probably need to create a JESD FSM class to handle that more generally.

Yes - agree - in fact the fsm stuff in this PR is a bit more advanced since it has some asserts which handle a undesired corner case (bug)

PR #268, opened 2021-11-15
Added a condition to skip the interface gain narrowband test - ADRV9002

HRosete:
# Description

Added a condition to skip the interface gain narrowband test if the baseband RX sample rate is greater than 1MHz. This answers the issue stated in [SDGSW-737](https://jira.analog.com/browse/SDGSW-737)

# How has this been tested?

Tested and verified in the Test Harness

- [x] [HW_pyadi_test #99](http://10.116.171.86:8080/blue/organizations/jenkins/HW_tests%2FHW_pyadi_test/detail/HW_pyadi_test/99/pipeline/583/)

**Test Configuration**:
* Hardware: Zed + ADRV9002, ZCU102 + ADRV9002
* OS: ADI-Kuiper-Linux

# Checklist:

### DIFF
--- a/test/test_adrv9002_p.py
+++ b/test/test_adrv9002_p.py
@@ -136,6 +136,20 @@ def test_adrv9002_str_attr(
 def test_adrv9002_interface_gain_narrowband(
     test_attribute_multipe_values_with_depends, iio_uri, classname, attr, depends, val
 ):
+    from adi.adrv9002 import adrv9002
+
+    sdr = adrv9002(iio_uri)
+    if attr == "interface_gain_chan0":
+        if sdr.rx0_sample_rate > 1000000:
+            pytest.skip(
+                "Baseband RX1 Sample Rate should be less than 1MHz to run this test."
+            )
+    elif attr == "interface_gain_chan1":
+        if sdr.rx1_sample_rate > 1000000:
+            pytest.skip(
+                "Baseband RX2 Sample Rate should be less than 1MHz to run this test."
+            )
+
     test_attribute_multipe_values_with_depends(
         iio_uri, classname, attr, depends, val, 0
     )
--- a/test/test_adrv9002_p.py
+++ b/test/test_adrv9002_p.py
@@ -15,6 +15,7 @@ lte_40_lvds_stream = profile_path + "lte_40_lvds_api_48_26_4.stream"
 lte_5_cmos_profile = profile_path + "lte_5_cmos_api_48_26_4.json"
 lte_5_cmos_stream = profile_path + "lte_5_cmos_api_48_26_4.stream"
 
+
 #########################################
 @pytest.mark.iio_hardware(hardware)
 @pytest.mark.parametrize("classname", [(classname)])


test/test_adrv9002_p.py
```diff
@@ -136,6 +136,12 @@ def test_adrv9002_str_attr(
 def test_adrv9002_interface_gain_narrowband(
     test_attribute_multipe_values_with_depends, iio_uri, classname, attr, depends, val
 ):
+    from adi.adrv9002 import adrv9002
+
+    sdr = adrv9002(iio_uri)
+    if sdr.rx0_sample_rate > 1000000:
```
nunojsa, line 142: We should further improve this. Even though this might work most of the time, for correctness, we test `sdr.rx0_sample_rate > 1000000` when running the test for RX1 and `sdr.rx1_sample_rate > 1000000` when running for RX2

nunojsa (changes requested):

test/test_adrv9002_p.py
HRosete: Oh, okay. Thanks for pointing that out. I'd just edit the conditional then.

HRosete: I've added the sdr.rx1_sample_rate property to be checked on the conditional. Please see if this will cater to both of the channels.

test/test_adrv9002_p.py
```diff
@@ -139,8 +139,10 @@ def test_adrv9002_interface_gain_narrowband(
     from adi.adrv9002 import adrv9002
 
     sdr = adrv9002(iio_uri)
-    if sdr.rx0_sample_rate > 1000000:
-        pytest.skip("Baseband RX Sample Rate should be less than 1MHz to run this test.")
+    if (sdr.rx0_sample_rate or sdr.rx1_sample_rate) > 1000000:
+        pytest.skip(
+            "Baseband RX Sample Rate should be less than 1MHz to run this test."
+        )
```
nunojsa, line 145: This is still not quite there... What I meant (even though unlikely) is that 1 channel can pass the test while the other might skip it... So, I think it should look something like:

```
if attr == "interface_gain_chan0":
    if sdr.rx0_sample_rate > 1000000:
        pytest.skip("some message")
else:
    if sdr.rx1_sample_rate > 1000000:
        pytest.skip("some message")
```

See the difference?

nunojsa (changes requested):
Take also a look you your commits... It looks like you're just adding commits on top of your branch which pollutes the history. Rebase your feature branch so that you just have two commits as in the original pull request.

test/test_adrv9002_p.py
HRosete: Hello, Nuno. I now get your point. I initially assumed that it would just automatically read the property from the channel it tests, not all of the device properties, but apparently, it's not the case. That's why I made the initial conditional to be very simple. I didn't know that I actually made a loophole for errors in the test. I'm sorry. I was mistaken. It does read all of the device properties at each test run, albeit it's testing an attribute in channel 1 or channel 2. Thank you.

HRosete: As for the commits, I also rebased them. We won't have any problems with a polluted history.

PR #299, opened 2022-02-04
Dev sync start

mhennerich:
# Description

Support for the new EXT_SYNC sync start HDL core feature

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/adi/sync_start.py
+++ b/adi/sync_start.py
@@ -1,0 +1,128 @@
+# Copyright (C) 2022 Analog Devices, Inc.
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification,
+# are permitted provided that the following conditions are met:
+#     - Redistributions of source code must retain the above copyright
+#       notice, this list of conditions and the following disclaimer.
+#     - Redistributions in binary form must reproduce the above copyright
+#       notice, this list of conditions and the following disclaimer in
+#       the documentation and/or other materials provided with the
+#       distribution.
+#     - Neither the name of Analog Devices, Inc. nor the names of its
+#       contributors may be used to endorse or promote products derived
+#       from this software without specific prior written permission.
+#     - The use of this software may or may not infringe the patent rights
+#       of one or more patent holders.  This license does not release you
+#       from the requirement that you obtain separate licenses from these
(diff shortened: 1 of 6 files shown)


tfcollins:
I added the new context here: https://github.com/tfcollins/pytest-libiio/tree/tfcollins/update-ad9081 eye_data breaks iio_attr. I wonder if that should be a scan element.

Anyway, there are some test failures with related to setting loopback modes. Current tests are hardcoded to 4 channels not 1.

adi/sync_start.py
```diff
@@ -0,0 +1,128 @@
...
+            effect only the EXT_SYNC synthesis parameter is set.
+
+        - trigger_manual: Writing this key will issue an external sync event
+            if it is hooked up inside the fabric. This key has an effect
+            only the EXT_SYNC synthesis parameter is set.
+            This key self clears.
+        """
+        try:
+            return self._get_iio_dev_attr_str("sync_start_enable", _ctrl=self._txdac)
+        except:  # noqa: E722
```
tfcollins, line 60: Can we populate the exceptions? I know ignoring them completely has caused issues in the past with debug.

mhennerich: This will except only in case sync_start_enable attribute doesn't exist. In this case we fall back to the hacky register read.
I did that for compatibility reasons ...

mhennerich:
> I added the new context here: https://github.com/tfcollins/pytest-libiio/tree/tfcollins/update-ad9081 eye_data breaks iio_attr. I wonder if that should be a scan element.

eye_data is a binary attribute attribute.
https://wiki.analog.com/resources/tools-software/linux-drivers/jesd204/axi_adxcvr#jesd204_eye_scan

tfcollins:
It causes the iio tools to hang. Maybe we should update them to better handle this?

Do we want a eye scan util in python too (eventually)?

mhennerich:
> It causes the iio tools to hang. Maybe we should update them to better handle this?

Not sure why it hangs, we have other binary sysfs as tributes as well. For example gain tables and profile configs.

> Do we want a eye scan util in python too (eventually)?

I think it would make sense.

PR #313, opened 2022-03-08
Adpd2140 get sensor

adrimbarean:
# Description

Add adpd1080 pyadi-iio class, basic example and gesture sensor/theremin example.

## Type of change

# How has this been tested?

Class and examples tested with hardware system. The first and second pins of the PMOD are not connected to the board by default in rev A and a wire needs to connect pin 1 of the PMOD to IO13 of the ADICUP3029.

- Test 1: Change the com in the main function of the theremin module and run it with Python 3. Wait a few seconds (about 5) to calibrate and place hand above the PMOD. After keeping it for up to 6 seconds move it away. The program should detect the way of the movement.

**Test Configuration**:
* Hardware: ADICUP3029 + CN0569
* OS: Windows (Linux should be compatible)

# Checklist:

- [x] I have made corresponding changes to the documentation
- [x] Any dependent changes have been merged and published in downstream modules


adi/adpd1080.py
```diff
@@ -0,0 +1,110 @@
...
+            name = ch._id
+            self._rx_channel_names.append(name)
+            self.channel.append(self._channel(self._ctrl, name))
+        rx.__init__(self)
+        self.rx_buffer_size = 16
+
+    def rx(self):
+            buff = super().rx()
+            del self._rx__rxbuf
+            return buff
```
tfcollins, line 79: Do you really need to delete the buffer between calls?

adi/adpd1080.py
```diff
@@ -0,0 +1,110 @@
...
+        def offset(self, value):
+            self._set_iio_attr(self.name, "offset", False, value)
+
+        @property
+        def sample_rate(self):
+            """Sets sampling frequency of the ADPD1080."""
+            return self._get_iio_attr(self.name, "sampling_frequency", False)
+
+        @sample_rate.setter
+        def sample_rate(self, value):
```
tfcollins, line 109: Is sample rate the same across all channels? If so can we move this up out of the individual channels and just have 1?

adrimbarean: Will do.

adi/adpd1080.py
adrimbarean: This was done because in an application we needed to read a few samples and then take the average and write it into the offset register of the device. But writes to the device did not work anymore after buffer reads, unless the buffer was deleted first.

adrimbarean:
V1:
 - move sample rate to device properties.

adrimbarean:
V2:
 - solved auto checking complaints (the remaining one is unrelated to this PR and solved [here](https://github.com/analogdevicesinc/pyadi-iio/pull/314))

buha (approved):

adi/adpd1080.py
tfcollins: Hmm that seems really strange. Is this a bug in the driver or the desired behavior?

adrimbarean:
V3:
 - rebase on current master.

adrimbarean:
Codacy complaints that there should be no blank lines before class docstring, but there are none. Seems like a false positive.

adi/adpd1080.py
tfcollins: I can leave it up to you and @scuciurean about if this is proper. Just let me know and we'll merge.

adrimbarean:
V4:
- remove unnecessary timeout assignation.

adrimbarean:
V5:
 - comment code needed for workaround; to be uncommented after workaround is done;
 - re-apply black.

examples/adpd1080.py
```diff
@@ -0,0 +1,46 @@
...
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+# THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import adi
+
+# Set up ADPD1080
+adpd1080 = adi.adpd1080(uri="serial:COM23")
+adpd1080.rx_buffer_size = 40
+# commented on adicup3029 platform, may work on others
+# todo: undomment when uart issue on adicup is fixed
```
scuciurean, line 40: Maybe we can remove this type of comments and add them as issues or tickets

scuciurean:
Apart from the 2 code analysis issues, it looks good to me.

adrimbarean:
> Apart from the 2 code analysis issues, it looks good to me.

These seem like false positives to me. The issue is that blank lines are not allowed before class docstring, but there are no blank lines here.

adrimbarean:
V6:
 - remove todo comments in favor of an issue to track workarounds;

examples/adpd1080.py
adrimbarean: [Done](https://github.com/analogdevicesinc/pyadi-iio/issues/321)

PR #382, opened 2022-10-24
Reorg rx buffering

tfcollins:
Reorganize RX buffer to not interleave and then deinterleave constantly

Signed-off-by: Travis F. Collins <travis.collins@analog.com>

### DIFF
--- a/adi/rx_tx.py
+++ b/adi/rx_tx.py
@@ -51,6 +51,7 @@ class phy(attribute):
 
 class rx_tx_common(attribute):
     """Common functions for RX and TX"""
+
     def _annotate(self, data, cnames: List[str], echans: List[int]):
         return {cnames[ec]: data[i] for i, ec in enumerate(echans)}
 
@@ -141,7 +142,7 @@ class rx(rx_tx_common):
         if not isinstance(value, list):
             raise Exception("rx_enabled_channels must be a list")
         if not all(isinstance(x, int) for x in value) and not all(
-                isinstance(x, str) for x in value
+            isinstance(x, str) for x in value
         ):
             raise Exception(
                 "rx_enabled_channels must be a list of integers or "
@@ -188,9 +189,7 @@ class rx(rx_tx_common):
         for i in self.rx_enabled_channels:
             v = self._rxadc.find_channel(self._rx_channel_names[i])
             if "scale" in v.attrs:
-                scale = self._get_iio_attr(
-                    self._rx_channel_names[i], "scale", False
-                )
+                scale = self._get_iio_attr(self._rx_channel_names[i], "scale", False)
             else:
                 scale = 1.0
             rx_scale.append(scale)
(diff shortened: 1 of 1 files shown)


adi/rx_tx.py
```diff
@@ -255,84 +252,62 @@ def __rx_unbuffered_data(self):
...
-            x[idx::stride] = np.bitwise_and(x[idx::stride], (1 << df.length) - 1)
-
-        if not self._rx_stack_interleaved and stride>1:
-            stacked_data = []
-            for idx, ec in enumerate(iio_enabled_channels):
-                stacked_data.extend(x[idx::stride])
-            x = stacked_data
-        return x
+            ecn = [self._rx_channel_names[m] for m in self.rx_enabled_channels]
+
```
adisuciu, line 275: enable channels need to be sorted by index ... if in user application you enable channels such as:
[ch2,ch0,ch1]
but the samples come interleaved as 
ch0,ch1,ch2,ch0,ch1.ch2

You will apply ch2 format to ch0 and so on. This is important for multitype.
This is implemented here: 
https://github.com/analogdevicesinc/pyadi-iio/blob/rx_tx-dynamic-channel-type-conversion/adi/rx_tx.py#L273
However my implementation makes use of iio_channels instead of iio channel names which makes things a little "dirtier"

adi/rx_tx.py
```diff
@@ -255,84 +252,62 @@ def __rx_unbuffered_data(self):
...
+        for name in ecn:
+            chan = self._rxadc.find_channel(name)
+            # create format string
+            df = chan.data_format
+            fmt = (">" if df.is_be is True else "<") + (
```
adisuciu, line 280: If you do per channel read, then you don't need to set endianness as channel read already does it for you. Setting this will convert endianness twice (once in frombuffer and once in channel read) which leads to incorrect results. 
This should be removed (">" if df.is_be is True else "<").

If we don't use channel read (but rather buffer.read() ) then endianness is necessary in frombuffer conversion.

adi/rx_tx.py
```diff
@@ -255,84 +252,62 @@ def __rx_unbuffered_data(self):
...
+                "i" if df.is_signed is True else "u"
+            )
+            fmt += str(df.length // 8)
+            ba = chan.read(self.__rxbuf)  # Do local type conversion
+            data_channel_interleaved.append(np.frombuffer(ba, dtype=fmt))
+
+        return data_channel_interleaved
```
adisuciu, line 287: So as I understand it, the default option is channel interleaved. I kinda missed this point, where I thought that sample interleaved is the default.

If the user wants sample interleaved they might as well do it in their user programs. I think this makes more sense!

adi/rx_tx.py
```diff
@@ -255,84 +252,62 @@ def __rx_unbuffered_data(self):
...
-            cplx = np.zeros(len(x) // stride, dtype=complex)
-            np.multiply(x[indx + 1::stride], 1j, cplx)
-            np.add(x[indx::stride], cplx, cplx)
-            sig.append(cplx)
-            indx = indx + 2
+        if len(x) % 2 != 0:
+            raise Exception(
+                "Complex data must have an even number of component channels"
+            )
+        out = [x[i] + 1j * x[i + 1] for i in range(0, len(x), 2)]
```
adisuciu, line 295: Nice reduction :)

adisuciu (changes requested):

adi/rx_tx.py
adisuciu: EDIT! Actually, nevermind ! If you read per channel first, you don't care about all this ! Just read channel using libiio API and apply it's own formatting. No changes needed here !

PR #396, opened 2023-01-03
Ad5592r

mthoren-adi:
# Description

Add AD5592r support and examples, including curve tracer used for workshop.

## Type of change

# How has this been tested?

- [X] Hardware testing using EVAL-AD5592R-PMDZ on Raspberry Pi 4 and Raspberry Pi 400
- [X] Hardware testing using EVAL-AD5593R (I2C variant) on same setup
- [X] Pytest attribute test passed

**Test Configuration**:
* Hardware: Raspberry Pi 4, 400
* OS: ADI Kuiper Linux, 2022 r2 RC1

# Documentation

# Checklist:

- [X] New and existing unit tests pass locally with my changes
- [X] Any dependent changes have been merged and published in downstream modules


mthoren-adi:
@tfcollins @adisuciu  curious your opinion on the channel layout:
https://github.com/mthoren-adi/pyadi-iio/blob/ad5592r/adi/ad5592r.py#L71:L85
Not super convenient for iterating, but how often do you need to blindly iterate in an actual application? This:
https://github.com/mthoren-adi/pyadi-iio/blob/ad5592r/examples/ad5592r_example.py#L48:L52
gets the job done, if it's not too wonky.
In a "real" application where you know exactly what channel is connected where, associating channels looks pretty intuitive:
https://github.com/mthoren-adi/pyadi-iio/blob/ad5592r/examples/ad5592r_examples/ad5592r_curve_tracer.py#L86:L91

tfcollins:
@mthoren-adi can you rebase onto master

mthoren-adi:
@tfcollins Rebased my branch, looks like there are conflicts with analogdevicesinc:ad5592r (expected?) Is this sufficient, or shall I open a PR directly to master?

tfcollins:
O missed that it wasn't going to master. But I would just point it to master and do the review that way

mthoren-adi:
@tfcollins Just opened a new PR from my ad5592r to analogdevicesinc/master.

adi/ad5592r.py
```diff
@@ -0,0 +1,132 @@
...
+
+        @property
+        # AD559XR channel temp offset value
+        def offset(self):
+            return self._get_iio_attr(self.name, "offset", self._output)
+
+        @raw.setter
+        def raw(self, value):
+            if self._output is True:
+                self._set_iio_attr(self.name, "raw", self._output, value)
```
tfcollins, line 114: can you add an exception here if its not an output

mthoren-adi: I set up separate channel types for ADC, DAC, and temperature. Adding an exception worked, but it seemed odd to have a setter there in the first place. With the new scheme, setting and getting DAC raw is as expected, attempting to write to an ADC raw gives:
AttributeError: can't set attribute

Also I realized that scale_avaliable doesn't apply to temperature, and offset doesn't apply to ADC or DAC, more reason for separate channel types.

Does this look reasonable?

adisuciu:
I feel like it should be possible to create a list/dictionary of channel names, which would make iteration easier. Does impementing with dynamic attributes give any advantages over smth like this ? (I didn't actually test this, as I don't have a board):


```
channel_names={}
  # Dynamically get channels after the index
        for ch in self._ctrl.channels:
            name = ch._id
            output = ch._output
            if name == "temp":
                #setattr(self, name, self._channel_temp(self._ctrl, name, output))
				channel_names[name+"_temp"].append( self._channel_temp(self._ctrl, name, output)
            else:
                if output is True:
                    setattr(
                        self, name + "_dac", self._channel_dac(self._ctrl, name, output)
                    )
					channel_names[name+"_dac"].append( self._channel_temp(self._ctrl, name, output)
                else:
                    setattr(
                        self, name + "_adc", self._channel_adc(self._ctrl, name, output)
                    )
					channel_names[name+"_adc"].append( self._channel_temp(self._ctrl, name, output)
```

You should then be able to
```
for ch in channel_names:
	print(ch.name)
```

but also
`Vcsense = my_ad5592r.channel_names["voltage1_adc"]`

I think that channel_names is a bad name, maybe channels_by_name ? I don't know .. 
Also this could be done by the rx/tx class, at the base level, not by each driver wrapper. 
I did NOT test this, but it should work. Any thoughts ? 

-Adrian

mthoren-adi:
@adisuciu I like the way this looks, so far the only idiosyncrasy is tabbing after accessing a channel's value doesn't show the attributes:
my_ad5592r.ch_names["voltage2_dac"].<<tab>>
Am I missing something?
And calling the dictionary "ch_names" seems okay, unless you or @tfcollins  have another suggestion.

adisuciu (approved):

tfcollins:
No readline (autocomplete) doesn't inspect indexed dictionary items.

Why not just use the property function to assign the adc/dac values as just properties?

adisuciu:
> Why not just use the property function to assign the adc/dac values as just properties?

My idea was that by implementing a dictionary you can iterate through them .. The iio context (channel IO configuration) looks different based on devicetree config.

mthoren-adi:
@tfcollins , @adisuciu,  Updated per recent discussion with Travis. The channel layout is described here:
https://github.com/mthoren-adi/pyadi-iio/blob/ad5592r/adi/ad5592r.py#L47
which looks like this in an example application:
https://github.com/mthoren-adi/pyadi-iio/blob/ad5592r/examples/ad5592r_examples/ad5592r_curve_tracer.py#L59

If it looks okay I'll squash the commits.

PR #419, opened 2023-03-16
Added pyadi-iio support for ad7690 device family

mariel-tinaco:
# Description

Added pyadi-iio module and supported source/document files for ad7690 device family

Fixes # (issue)

## Type of change

# How has this been tested?

- Since the IIO driver support for the ad7690 family is yet to be available, this was tested on the EVAL-AD7689 evaluation board which is a multi-channel version of the ad7690 family. We used a Raspberry Pi 4B with custom device tree and Linux driver support for AD7949 (supports ad7949, ad7682, ad7689)

**Test Configuration**:
* Hardware: EVAL-AD7689EDZ and Raspberry Pi 4B
* OS: Windows 10

# Documentation

https://www.analog.com/en/products/ad7690.html

# Checklist:

- [x] I have made corresponding changes to the documentation


tfcollins:
This new class is very very similar to the AD7689 implementation. It really just has a flat rather than nested structure. The AD7689 looks like it could directly handle the AD7690 configuration, minus the unbuffered prop.

What is the reasoning for not just using the AD7689 class?

Exposing the buffering control mode is not something we usually do. Why was that added?

MarielTinaco:
Thanks for the comment. We were also debating whether to just add the ad7690 device family to the compatible parts of ad7689. I was worried about modifying the existing support for ad7689. 

As for the reasoning, The ad7690 device family would no longer have the need for the channel class since it is a single-channel unlike the ad7689 which is multi-channel. so the raw and scale prop would be device attributes instead of device channel attributes. Also we are unable to perform hardware tests for ad7690 due to the lack of IIO driver support. So we had to implement support for ad7689 as well.

As for the buffering control mode, the AD7689 doesn't enable buffered rx channel, as checked with the hardware test. Data can only be read via unbuffered rx.

mphalke:
@MarielTinaco @tfcollins I did check the linux drivers for ad7689 and seems like there is no buffer support added yet. These drivers are very old so may be the reason.
https://github.com/analogdevicesinc/linux/blob/master/drivers/iio/adc/ad7949.c

We had been working on the Tinyiiod based application (non-linux) and for which I had to add buffer support in the ad7689 class. Hope this is fine?
https://github.com/analogdevicesinc/precision-converters-firmware/tree/main/projects/ad7689_iio
https://github.com/analogdevicesinc/precision-converters-firmware/blob/main/projects/ad7689_iio/scripts/ad7689_data_capture.py

tfcollins:
It is possible to check if there are buffer capable devices. Maybe that would be a better path. You would do something like this:
```python
self._rx_unbuffered_data = len(self._rxadc.buffer_attrs) == 0
```

PR #432, opened 2023-04-19
[FIX] Correct test validation for attr tests

tfcollins:
# Description

PR fixes core test bug that checks values within tolerances

Fixes # (issue)

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# How has this been tested?

- [ ] All AD9364 based tests

**Test Configuration**:
* Hardware: ADRV9364+BOB
* OS: Ubuntu 22.04

# Documentation

# Checklist:

- [X] New and existing unit tests pass locally with my changes
- [X] Any dependent changes have been merged and published in downstream modules


mthoren-adi:
@tfcollins  Okay I have something running for my readonly case (with the dev_interface_sub_channel return condition flipped as noted above), and with the equivalent of:
https://github.com/analogdevicesinc/pyadi-iio/blob/98cc4e349cfd9d9598cd95308b4b9325ad91697a/test/attr_tests.py#L67
removed in my attribute_single_value_readonly test. (I think this should be removed a couple of other places in attr_tests, take a look.)

I can zoom in on expected scale values (which are constant), setting an upper and lower limit, and the test passes and fails as expected when I purposely move the window so the value falls outside. Works for the float voltage scale value, and what looks like an integer temp scale value. 

Would it be possible to add a readonly = False default argument here:
https://github.com/analogdevicesinc/pyadi-iio/blob/98cc4e349cfd9d9598cd95308b4b9325ad91697a/test/common.py#L140
And then make this conditional:
https://github.com/analogdevicesinc/pyadi-iio/blob/98cc4e349cfd9d9598cd95308b4b9325ad91697a/test/common.py#L154

if readonly is False:
        setattr(getattr(sdr, sub_channel), attr, val)

?
(This could also be applied to dev_interface (no subchannel.))

This should pass your AD9364 tests above. I don't mean to be lazy, but it would be good to know it's tested exhaustively. Then I'll add in a attribute_single_value_readonly test in my AD7291 commit.

test/attr_tests.py
```diff
@@ -67,7 +67,7 @@ def attribute_single_value(
                 <= tol
```
mthoren-adi, line 67: Missed one...

PR #449, opened 2023-06-30
rx_tx: Add tx_data_type

SaikiranGudla:
To enable buffer transmission with data types larger than 16 bits.

This change extends support to transmit data to DACs with more than 16-bit precision.


SaikiranGudla:
May I know if there are any comments from the reviewers?

nunojsa:
> May I know if there are any comments from the reviewers?

You should tag them so that this is not missed :)

PR #490, opened 2023-10-25
adi/ad908[1|4]_mc.py: Fix label parsing in _map_to_dict()

mhennerich:
# Description

In recent kernels the IIO core adds labels for all channels with extended names. This now causes undesired behavior. Before splitting, check if the split string is part of the string. This will also catch the 'buffer_only' case.

## Type of change

# How has this been tested?

On HW using Linux 6.1

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Hardware:
* OS:

# Documentation

### DIFF
--- a/adi/ad9081_mc.py
+++ b/adi/ad9081_mc.py
@@ -13,7 +13,7 @@ from adi.sync_start import sync_start
 
 
 def _map_to_dict(paths, ch, dev_name):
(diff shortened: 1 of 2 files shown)


tfcollins (commented):
Change is fine. Think we can get the updated context files?

PR #518, opened 2024-01-18
Eye scan support

tfcollins:
Add JESD eye scan helper functions and add JESD debug support to AD9081.

### DIFF
--- a/adi/jesd_internal.py
+++ b/adi/jesd_internal.py
@@ -2,12 +2,10 @@
 #
 # SPDX short identifier: ADIBSD
 
-from abc import ABCMeta, abstractmethod
-
 from .sshfs import sshfs
 
 
-class eye_scan(metaclass=ABCMeta):
+class eye_scan(object):
     _jesd_es_duration_ms = 100
     _jesd_prbs = 7
 
--- a/adi/jesd_internal.py
+++ b/adi/jesd_internal.py
@@ -2,10 +2,99 @@
 #
 # SPDX short identifier: ADIBSD
 
+from abc import ABCMeta, abstractmethod
+
 from .sshfs import sshfs
 
 
-class jesd:
+class eye_scan(metaclass=ABCMeta):
+    _jesd_es_duration_ms = 100
+    _jesd_prbs = 7
+
+    _half_rate = {"mode": "Half Fate", "scale": 0.004}
+    _quarter_rate = {"mode": "Quarter Rate", "scale": 0.001}
+
+    lanes = {}
+
+    def get_eye_data(self, lanes=None):
+        """Get JESD204 eye scan data
+
+        Args:
+            lanes (list, optional): List of lanes to get data for. Defaults to None which will get data for all lanes.
+
+        Returns:
+            dict: Dictionary of lane data. Keys are lane numbers, values are dictionaries with keys "x", "y1", "y2", and "mode".
+                where "x" is the x-axis data SPO, "y1" is the y-axis data for the first eye, "y2" is the y-axis data for the second eye,
+                in volts
+
+        """
+        # Check if supported
+        if "bist_2d_eyescan_jrx" not in self._ctrl.debug_attrs:
+            raise Exception("2D eye scan not supported on platform")
+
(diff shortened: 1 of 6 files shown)


adi/jesd_internal.py
```diff
@@ -5,7 +5,94 @@
...
+class eye_scan(object):
+    _jesd_es_duration_ms = 100
+    _jesd_prbs = 7
+
+    _half_rate = {"mode": "Half Fate", "scale": 0.004}
+    _quarter_rate = {"mode": "Quarter Rate", "scale": 0.001}
+
+    lanes = {}
+
+    def get_eye_data(self, lanes=None):
```
mhennerich, line 17: Not sure if this belongs into jesd_internal 
This is a MxFE side JRX eye_scan and is specific to MxFE...
I would assume this should go into ad9081.py?

adi/jesd_internal.py
```diff
@@ -5,7 +5,94 @@
 from .sshfs import sshfs
 
 
-class jesd:
+class eye_scan(object):
+    _jesd_es_duration_ms = 100
+    _jesd_prbs = 7
+
+    _half_rate = {"mode": "Half Fate", "scale": 0.004}
```
mhennerich, line 12: Let's scale to mV for half rate the API already scales things.
So here we use 1.0

adi/jesd_internal.py
```diff
@@ -5,7 +5,94 @@
...
+    _quarter_rate = {"mode": "Quarter Rate", "scale": 0.001}
```
mhennerich, line 13: For quarter rate we need to scale until the API updates the next time.
So scale should be 4.0 here

adi/jesd_internal.py
```diff
@@ -5,7 +5,94 @@
...
+                else:
+                    spo = [int(x) for x in eye_line.split(",")]
+                    x.append(spo[0])
+                    y1.append(spo[1] * scale)
+                    y2.append(spo[2] * scale)
+
+            graph_helpers = {
+                "xlim": [-info[1] / 2, info[1] / 2 - 1],
+                "xlabel": "SPO",
+                "ylabel": "EYE Voltage (V)",
```
mhennerich, line 79: Let's do mV instead

examples/ad9081_jesd_eye_diagram.py
```diff
@@ -0,0 +1,40 @@
...
+# Configure properties
+print("--Setting up chip")
+
+dev._ctx.set_timeout(90000)
+
+fig = plt.figure()
+
+eye_data_per_lane = dev._jesd.get_eye_data()
+num_lanes = len(eye_data_per_lane.keys())
+
```
mhennerich, line 18: I updated the driver to return -EINVAL if a physical lane is not used (virtually mapped)

So we could do:

```
def get_mapped_lanes(lanes):
    mapped_lanes = []
    for lane in lanes:
        try:
            dev._ctrl.debug_attrs["bist_2d_eyescan_jrx"].value = f"{lane} 7 1"
            mapped_lanes.append(lane)
        except:
            continue
    return mapped_lanes

all_lanes = ['0', '1', '2', '3', '4', '5', '6', '7']
lanes = get_mapped_lanes(all_lanes)

```

adi/jesd_internal.py
tfcollins: I reorganized the inheritance so this will be only pulled in by MxFE

adi/jesd_internal.py
tfcollins: Fixed

adi/jesd_internal.py
tfcollins: Fixed

adi/jesd_internal.py
tfcollins: Fixed

examples/ad9081_jesd_eye_diagram.py
tfcollins: I added some pieces around this. There was some confusion around the link naming in sysfs which always goes 0,1,2,... N which is not the actual lanes mapped index. This is handled now

PR #541, opened 2024-03-11
Enhance ADRV9009-ZU11EG multi-class to support selective FMComms8 suport

tfcollins:


cafeclimber:
Hey Travis. Sorry for the delay, but finally got around to testing this today. Everything seems to work! I did have to make a few edits to the multi class. There are some checks for fmcomms8 that need to be changed to index into the list. Specifically lines `234`, `246`, `254`, `271`, `360`, and `374`. I just did like you did earlier and changed the for loop to be (e.g.)

```python
for i, dev in enumerate([self.primary] + self.secondaries):
    dev._clock_chip.attrs["sleep_request"].value = "1"
    if self.fmcomms8[i]:
        dev._clock_chip_fmc.attrs["sleep_request"].value = "1"
    dev._clock_chip_carrier.attrs["sleep_request"].value = "1"
```

PR #584, opened 2024-07-29
Fixes for adrv9371 and adrv9009

tagoylo:
# Description

This PR includes fixes for adrv9371 and adrv9009 as well as their tests.

- adrv9371 changes includes checking if decimation is supported before setting or getting sampling frequency. It was failing on arria10 since it is not supported there.
- adrv9009 changes includes update on the obs lo attribute name. Also added new profile (got it from latest osc). 

## Type of change

# How has this been tested?

- [x] Tested on hardware setup. arria10+adrv9371 and arria10+adrv9009

**Test Configuration**:
* Hardware:
* OS:

# Documentation

# Checklist:


tagoylo:
@tfcollins PR is ready for review. I updated the emulated test workflow - to run the test without the optional dependency.

adi/ad937x.py
```diff
@@ -170,44 +170,60 @@ def tx_rf_bandwidth(self):
...
-        return val == avail[1]
+        avail = None
+        try:
+            avail = self._get_iio_attr_str(
+                "voltage0_i", "sampling_frequency_available", False, self._rxadc
+            )
+            avail = avail.strip().split(" ")
+            return val == avail[1]
+        except KeyError:
+            return val == avail
```
tfcollins, line 184: This applies to the other blocks that share this logic.

1. If sampling_frequency_available does not exist this will generate a strange error. Something like sampling_frequency_available is not a valid key, which won't make any sense. So there needs to be an except catch for that case
2. The case when it does but only 1 value exists doesn't really make sense or the logic flow is convoluted. If only 1 value exists then it should always return false.

tfcollins (changes requested):
Primarily just questions on the DEC8/INT8 updates

.github/workflows/test.yml
```diff
@@ -49,9 +49,10 @@ jobs:
           bash ./.github/scripts/install_pydeps.sh
           sudo apt install -y python3-tk
 
-      - name: Test
+      - name: Test without optional dependencies
         run: |
-                pytest -vs --cov=adi --scan-verbose --emu --junitxml="results.xml" -k 'not prod'
+                pip uninstall -y paramiko
+                pytest -vs --scan-verbose --emu --junitxml="results.xml" -k 'not prod'
```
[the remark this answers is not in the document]
tagoylo: Added the cov flag

adi/ad937x.py
tagoylo: I updated the implementation. It checks first if sampling_frequency_available exists then returns False if not. It it exists and only 1 value is available, it sets to False. And if it has 2 values available it returns True.

tfcollins (commented):
Is this really necessary https://github.com/analogdevicesinc/pyadi-iio/pull/584/commits/aea0aca274ecd6371bfcbc1e8c914bb6acc21e08 ? Did something break these?

tagoylo:
Yes, all of the tests were skipped before and were not run, and when I tried to run in emulation mode they are failing as it is trying to set some configs so I picked valid tests for emu.

PR #606, opened 2024-10-15
Update TX OBS to use new complex data flags

tfcollins:
Fix issue related to new complex data configuration flags for obs TX path.


jpineda3 (approved):
adrv9002 hw tests are now passing.

PR #615, opened 2024-11-19
Added support to AD4113 device

granquet:
1) Updated ad717x.py to include AD4113
2) Updated supported_parts.md file to add AD4113


tfcollins:
Can you provide a context emulation file if you have the hardware? It helps out a lot with test

granquet:
> Can you provide a context emulation file if you have the hardware? It helps out a lot with test

I do not have the hardware sadly.
I'm not sure we have one 4113 at baylibre.

I can probably infer an emulation xml file from another device in the 7193 family by looking at the differences in the driver ? if that can be helpful ?

tfcollins:
> > Can you provide a context emulation file if you have the hardware? It helps out a lot with test
> 
> I do not have the hardware sadly. I'm not sure we have one 4113 at baylibre.
> 
> I can probably infer an emulation xml file from another device in the 7193 family by looking at the differences in the driver ? if that can be helpful ?

No its fine. Thanks for the contribution

PR #644, opened 2025-02-28
context_manager.py: Fix error handling for usb

bia1708:
# Description

In the case of usb connections, no two iio.Context objects can exist at the same time. When the `dev_interface` functions are called inside the test functions, the function may fail before the `del sdr` operation can be performed and the iio.Context object is not destroyed. The next test that calls the `dev_interface` function will therefore also fail (since it cannot create another iio.Context object).
By transforming the `context_manager` class into a Pythonic context manager, we make use of the `with` statement. The `with` block will act as a wrapper over the usage of iio.Context objects and it will ensure that when errors are thrown, they are logged and memory is properly freed.
`context_manager` objects can still be created without the `with` statement (introduced change doesn't affect pre-existing code).

## Type of change

# How has this been tested?

Tested by running all pluto tests with a usb connection. The attribute test for setting the sampling rate (`test_pluto_attr[attr=sample_rate-start=2084000-stop=61440000-step=1-tol=4-repeats=100-classname=adi.Pluto]`) fails randomly. Whichever test comes after it (in my case it was `test_pluto_rx_data`) would fail with `Exception: No device found` because pyadi couldn't create a new iio.Context object. After implementing this fix, exceptions are caught and printed correctly and the memory for the iio.Context object is freed.

**Test Configuration**:
* Hardware: pluto
* OS: ubuntu22.04

# Documentation

# Checklist:

### DIFF
--- a/adi/context_manager.py
+++ b/adi/context_manager.py
@@ -36,3 +36,16 @@ class context_manager(object):
                 self._ctx = iio.Context(self.uri)
         except BaseException:
             raise Exception("No device found")
+
+    def close(self):
+        """Close the IIO context"""
+        if self._ctx:
+            del self._ctx
+            self._ctx = None
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.close()
+        return False
--- a/test/common.py
+++ b/test/common.py
@@ -106,32 +106,30 @@ def pytest_generate_tests(metafunc):
 def dev_interface(
     uri, classname, val, attr, tol, sub_channel=None, sleep=0, readonly=False
 ):
-    sdr = eval(classname + "(uri='" + uri + "')")
-    # Check hardware
-    if not hasattr(sdr, attr):
-        raise AttributeError(attr + " not defined in " + classname)
+    with eval(classname + "(uri='" + uri + "')") as sdr:
+        # Check hardware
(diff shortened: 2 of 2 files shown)


SRaus (approved):

test/common.py
jpineda3, line 1: Why is https://github.com/analogdevicesinc/pyadi-iio/blob/80796c55752931c2b8464168805f78f3e4e0c4e5/test/common.py#L206 not modified to us the with statement?

bia1708: because that `eval` instantiates an `iio.Device` object, not a `context_manager` object. If we want to be able to instantiate it using `with`, we would need to modify the logic of this function a bit

adi/context_manager.py
```diff
@@ -36,3 +36,19 @@ def __init__(self, uri="", _device_name=""):
...
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.close()
+        if exc_type:
+            print(f"Exception caught: {exc_value}")
+            return True
+        return False
```
tfcollins, line 54: You probably just want to replace these with "return False" so exceptions propagate normally. Library modules shouldn't print like this.

tfcollins (changes requested):
Just need a small change to the __exit__ method to not print

tfcollins (approved):

PR #660, opened 2025-04-10
pre-commit: update black, black-docs version

gastmaier:
# Description

typed-ast is not maintained anymore, and builds with newer python versions raises:

  ast27/Custom/../Include/compile.h:12:12: error: unknown type name ‘PyFutureFeatures’
     12 | PyAPI_FUNC(PyFutureFeatures *) PyFuture_FromAST(struct _mod *, const char *);

The minimum black version with 966baaa is 21.4b0,
but due to black-docs min version with the same fix (a24ed2d, 1.13.0), set as 22.12.0.

Fixes # (issue)

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How has this been tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Hardware:
* OS:

# Documentation

If this is a new feature or example please mention or link any documentation. All new hardware interface classes require documentation.

# Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have signed off all commits and they contain "Signed-off by: <insert name>"
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #662, opened 2025-04-15
pre-commit: Replace black, black-docs, isort with ruff

gastmaier:
# Description

Follow-up to #660, as a draft because we probably need to further tune the rules to execute `[tool.ruff.lint] select = [...]`.
Not a fan of jumping between tools, but I am glad that ruff can be installed from pip as usual.
This also changes the test.yml to run the checkers only on touched files (limited to 2048 commits (GH context window)).
In the case of force-push or new-branch-push event, it will check only the 6 top commits (arbitrary value, can be changed), and the environment variable FETCH_DEPTH_FALLBACK=true is also set.
The latter is used to fallback to check all files.
Another minor change is logging git diff with --no-pager instead of `> ; cat`.
I don't think we can 100% match the previous format checker.

## Type of change

# How has this been tested?

- [X] CI run on private pyadi-iio fork

**Test Configuration**:
* OS: OpenSuse Tumbleweed

# Documentation

Relevant docs:
https://docs.astral.sh/ruff/settings/#lint_select
https://docs.astral.sh/ruff/settings/#lintisort

# Checklist:

- [?] My code follows the style guidelines of this project
- [?] My changes generate no new warnings


gastmaier:
Hum. remembered that the github does not do a rebase for pull-requests, so getting the files names as I did is not accurate.
I will wait an initial review before further exploring.
I did my own [checkout](https://github.com/analogdevicesinc/doctools/tree/v1/checkout) to resolve this for (future-)linux ci, (since over there we are interested in checking each commit individually, as a patch series)
Let me know if we should all check all files to not have to deal with this, or rebase to ensure the commits are on top.

tfcollins:
Do you want to address the new lint issues in this PR?

gastmaier:
I believe we first need to decide if the provided format config meets the expectations by checking out the branch and running pre-commit.
If the config does make sense, we can do a single commit on top, right?

PR #674, opened 2025-07-17
Ad9084 misc sync

mhennerich:
# Description

- Added a new example script for AD9084/AD9082 synchronization and phase measurement, including full documentation and helper functions for signal analysis.
- Implemented the hmc7044 IIO device interface with dynamic channel property support and comprehensive docstrings.
- Implemented the adf4030 IIO device interface, supporting dynamic channel properties and temperature sensor access.
- Added the axi_aion_trig IIO device interface, dynamically generating trigger channel properties for flexible control.

## Type of change

# Checklist:


adi/sync_start.py
```diff
@@ -55,6 +55,50 @@ def tx_sync_start_available(self):
         except:  # noqa: E722
             return "arm"
 
+    @property
```
tfcollins, line 58: The B side methods won't be usable with non-Apollo ICs. Can we split these into a separate class then use inheritance to get the feature over to the AD9084 class. So look like:

```python

class sync_start_b(sync_start): # New class
    def tx_b_sync_start(self):
        ...

class ad9084(rx_tx, context_manager, sync_start_b): # updated
    ...



```

adi/sync_start.py
```diff
@@ -99,3 +143,48 @@ def rx_sync_start_available(self):
...
+                self._rxadc2.reg_write(0x48, 0x2)
+
+    @property
+    def rx_b_sync_start_available(self):
+        """ rx_sync_start_available: Returns a list of possible keys used for rx_sync_start """
+        try:
+            return self._get_iio_dev_attr_str(
+                "sync_start_enable_available", _ctrl=self._rxadc2
+            )
+        except:  # noqa: E722
```
tfcollins, line 189: This tends to be really error prone. Can we put the actual exception that is expected?

adi/sync_start.py
```diff
@@ -55,6 +55,50 @@ def tx_sync_start_available(self):
...
+              only the EXT_SYNC synthesis parameter is set.
+              This key self clears.
+        """
+        try:
+            return self._get_iio_dev_attr_str("sync_start_enable", _ctrl=self._txdac2)
+        except:  # noqa: E722
+            if self._txdac2.reg_read(0x80000068) & 0x1 == 0x1:
+                return "arm"
+            else:
+                return "disarm"
```
gastmaier, line 82: Something like this is better to handle the backwards compatbility instead of try catch
```
    if "sync_start_enable" in self._ctrl.attrs:
            return self._get_iio_dev_attr_str("sync_start_enable", _ctrl=self._txdac2)
    else:
            if self._txdac2.reg_read(0x80000068) & 0x1 == 0x1:
                return "arm"
            else:
                return "disarm"
```

gastmaier (commented):
on the try catch

adi/sync_start.py
mhennerich: done

PR #742, opened 2026-05-05
Dev/admt4000

kister-jimenez:
# Description

This pull requests adds support for [ADMT4000](https://www.analog.com/en/products/admt4000.html)

## Type of change

- [ ] Bug fix
- [x] New device class interface
- [ ] New feature on an existing class
- [ ] Breaking change
- [ ] Documentation only
- [ ] Test or CI only

# How has this been tested?

State explicitly whether this was tested against **real hardware**, an **emulated context** (iio-emu), or **not run**. Name the part(s) and the context URI or board used.

- Hardware / emulation: hardware
- Commands run (e.g. `python3 -m pytest -k ad4080 --adi-hw-map`): 
   `pytest -k test_admt4000 --uri=serial:COM4,230400 -v -s --hw='admt4000'`
   `pytest --emu --emu-xml=test/emu/devices/admt4000.xml -sv --hw="admt4000" --iio-coverage --cov=adi.admt4000 --skip-scan -k test_admt4000.py`
   `pytest --emu --emu-xml=test/emu/devices/admt4000ard1z.xml -sv --hw="admt4000ard1z" --iio-coverage --cov=adi.admt4000ard1z --skip-scan -k test_admt4000ard1z.py`

- Result:
- 

**Test configuration**
* Kernel / libiio version: libiio v0.26
* OS: no-os
* FPGA carrier (if applicable): N/A

# Documentation

- [ ] No documentation change needed
- [x] Added a new device page and linked it from the relevant toctree

# New device class interfaces

- [x] Class extends one of the common base classes in `adi.device_base` (`rx_chan_comp`, `tx_chan_comp`, or a `_no_buff` variant). If not, explain why in the description. See the [Device Base Classes](https://analogdevicesinc.github.io/pyadi-iio/dev/device_base.html) developer doc page.
- [x] `compatible_parts` lists every part number this class supports
- [x] Part numbers added to `supported_parts.md` (verify with `invoke checkparts`)
- [x] Emulation context (xml_gen, not iio_genxml) added under `test/emu/` and referenced from a test
- [x] Tests added that run against the emulation context in CI

# Checklist

- [x] `invoke precommit` passes locally
- [x] All commits are signed off (`Signed-off-by: Name <email>`)
- [ ] No new warnings from the build or doc generation
- [ ] Dependent changes in libiio, kernel, or HDL are merged and referenced


adi/tmc5240.py
```diff
@@ -0,0 +1,239 @@
...
+    def direction(self, value):
+        if value not in [1, -1]:
+            raise ValueError(
+                "Direction must be 1 (clockwise) or -1 (counter-clockwise)"
+            )
+
+        if value == -1:
+            value = 0
+
+        self._set_iio_dev_attr("direction", value, self._mot_ctrl)
```
mrcsosa, line 159: for consistency on all other device attribute setters, maybe use _set_iio_dev_attr_str across all setter attributes

adi/tmc5240.py
```diff
@@ -0,0 +1,239 @@
...
+            self._set_iio_attr(self.name, "calibscale", self._output, float(value))
+
+        @property
+        def processed(self):
+            """TMC5240 channel processed value."""
+            return self.raw * self.scale * self.calibscale
+
+        @processed.setter
+        def processed(self, value):
+            self.raw = int(value / (self.scale * self.calibscale))
```
mrcsosa: for a 0 divisor input, let's add a division by zero guard here, something like
divisor = self.scale * self.calibscale
if divisor == 0:
         raise ValueError("Cannot set processed: scale * calibscale is 0")

adi/tmc5240.py
```diff
@@ -0,0 +1,239 @@
...
+
+    class _mot_acceleration_channel(_mot_channel):
+        """TMC5240 acceleration channel."""
+
+        @property
```
mrcsosa, line 209: Uses _get_iio_attr here but _mot_channel uses _get_iio_attr_str for scale/calibscale.

adi/admt4000.py
```diff
@@ -0,0 +1,424 @@
...
+        self.sine = self._channel(self._ctrl, "sine")
+        self.cosine = self._channel(self._ctrl, "cosine")
+        self.radius = self._scaled_channel(self._ctrl, "radius")
+        self.gpio = self._gpio_accessor(self._ctrl)
+
+        if not self.disable_trigger:
+            if not trigger_name:
+                trigger_name = self._device_name + "-dev0"
+            self._trigger = self._ctx.find_device(trigger_name)
+            self._rxadc._set_trigger(self._trigger)
```
mrcsosa: find_device() returns None if the trigger is not found, this might cause _set_trigger(None) to fail with a unclear error

adi/tmc5240.py
```diff
@@ -0,0 +1,239 @@
...
+        def raw(self):
```
mrcsosa, line 210: Consider adding bound checks to validate the setter inputs

adi/tmc5240.py
kister-jimenez: I'll be using the API without the _str instead. I'll let pyadi-wrappers for this handle the casting.

adi/tmc5240.py
kister-jimenez: I'll add a Division by zero error instead.

adi/tmc5240.py
kister-jimenez: I'll retain the calls without the _str.

adi/admt4000.py
kister-jimenez: Will add a check.

adi/tmc5240.py
kister-jimenez: I'll let the firmware handle the bounds check.

adi/admt4000.py
```diff
@@ -0,0 +1,425 @@
...
+        """ADMT4000 1st, 2nd, 3rd, and 8th harmonic correction values. Passed as list of tuples of (magnitude, phase)."""
+        harmonics_raw = self._get_iio_dev_attr("harmonics", self._ctrl)
+        harmonics_mag_scale = self._get_iio_dev_attr(
+            "harmonics_magnitude_scale", self._ctrl
+        )
+        harmonics_phase_scale = self._get_iio_dev_attr(
+            "harmonics_phase_scale", self._ctrl
+        )
+        harmonics_values = []
+        values = harmonics_raw.replace(",", " ").split()
```
jpineda3: Isn't harmonics val already space-separated?

adi/admt4000.py
```diff
@@ -0,0 +1,425 @@
...
+            return self._get_iio_attr(self.name, "scale", False, self._ctrl)
+
+        @property
+        def processed(self):
+            """ADMT4000 channel processed value."""
+            scale = 1
+
+            try:
+                scale = self.scale
+            except (KeyError, AttributeError):
```
jpineda3, line 245: processed will still return a value even if reading scale (or offset for temp) fails. While the resulting value may be obviously incorrect, relying on that is could lead to subtle bugs if not caught downstream.

I think it would be better to let the exception propagate (or explicitly raise one) instead of catching it and returning an invalid value. This makes the failure explicit and easier to handle correctly at the call site.

adi/admt4000.py
kister-jimenez: removed the replace().

adi/admt4000.py
kister-jimenez: raise an error to propagate the error.

kister-jimenez:
V3:

- Raise exception in the processed properties
- removed redundant replace() before splitting harmonic values.
- minor code enhancements.

kister-jimenez:
V4:

- Added tests with high context-coverage for new classes.
- Added separate XMLs for each classes.


Context Coverage workflow has not pulled the latest pylibiio-test resulting to attribute calls not getting counted.

tfcollins:
> V4:
> 
>     * Added tests with high context-coverage for new classes.
> 
>     * Added separate XMLs for each classes.
> 
> 
> Context Coverage workflow has not pulled the latest pylibiio-test resulting to attribute calls not getting counted.

Reran the CI job and only admt4000ard1z is missing the mark

kister-jimenez:
> > V4:
> > ```
> > * Added tests with high context-coverage for new classes.
> > 
> > * Added separate XMLs for each classes.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Context Coverage workflow has not pulled the latest pylibiio-test resulting to attribute calls not getting counted.
> 
> Reran the CI job and only admt4000ard1z is missing the mark

I updated the test.

Is there a way to see the % values locally? I tried reading the IIO coverage report.md after running the test locally and I only got these:

```
# IIO Coverage Report

| System | Attribute Type | Coverage (%) |
|--------|----------------|----------|
| admt4000ard1z  | device_coverage | 2.4545454545454546 |
| admt4000ard1z  | channel_coverage | 5.714285714285714 |
| admt4000ard1z  | total_coverage | 3.7222222222222223 |
```

tfcollins (approved):

PR #770, opened 2026-06-23
ad9213 dev pyadi

TudorAndreiTicudean:
# Description

Add PyADI support for ad9213 based on already-existing: https://github.com/analogdevicesinc/pyadi-iio/tree/ad9213_dev

Fixes # (issue)

## Type of change

- [ ] Bug fix
- [X] New device class interface
- [ ] New feature on an existing class
- [ ] Breaking change
- [ ] Documentation only
- [ ] Test or CI only

# How has this been tested?

State explicitly whether this was tested against **real hardware**, an **emulated context** (iio-emu), or **not run**. Name the part(s) and the context URI or board used.

- Hardware / emulation: Real hardware - AD9213 on VCU118. Also verified against emulated context via iio-emu with ad9213.xml.
- Commands run (e.g. `python3 -m pytest -k ad4080 --adi-hw-map`):
-- pytest test/test_ad9213.py -v
-- python3 examples/ad9213_example.py
- Result: Test passed. Example ran successfully, register read/write and RX buffer capture confirmed working.

**Test configuration**
* Kernel / libiio version: 6.12.0-26956-g8edd67c2e728 / libiio 0.25
* OS: Linux buildroot (target), WSL2 Ubuntu (host)
* FPGA carrier (if applicable): VCU118

# Documentation

- [X] No documentation change needed
- [ ] Added a new device page and linked it from the relevant toctree

# New device class interfaces

- [X] Class extends one of the common base classes in `adi.device_base` (`rx_chan_comp`, `tx_chan_comp`, or a `_no_buff` variant). If not, explain why in the description. See the [Device Base Classes](https://analogdevicesinc.github.io/pyadi-iio/dev/device_base.html) developer doc page.
- [X] `compatible_parts` lists every part number this class supports
- [X] Part numbers added to `supported_parts.md` (verify with `invoke checkparts`)
- [X] Emulation context (xml_gen, not iio_genxml) added under `test/emu/` and referenced from a test
- [X] Tests added that run against the emulation context in CI

# Checklist

- [X] `invoke precommit` passes locally
- [X] All commits are signed off (`Signed-off-by: Name <email>`)
- [X] No new warnings from the build or doc generation
- [X] Dependent changes in libiio, kernel, or HDL are merged and referenced


tfcollins:
The driver naming here of `ad9213` is non-standard. We always use a convention `axi-<part>-rx-hpc`. I guess this is a devicetree naming bug?

tfcollins (commented):
The driver naming seems off. Please consult with the Linux team here.

adi/ad9213.py
```diff
@@ -3,14 +3,35 @@
 # SPDX short identifier: ADIBSD
 
 from adi.context_manager import context_manager
-from adi.rx_tx import rx_def
+from adi.rx_tx import rx
```
tfcollins: Keep rx_def. It has more supported features

adi/ad9213.py
```diff
@@ -3,14 +3,35 @@
...
+        self._ctrl = self._rxadc
+
+        rx.__init__(self)
+
+    @property
+    def sampling_frequency(self):
+        """sampling_frequency: Sample rate of the ADC in Hz."""
+        return self._get_iio_attr("voltage0", "sampling_frequency", False, self._rxadc)
+
+    def ad9213_register_read(self, reg):
```
tfcollins: this should be called just register_read. We already know its a ad9213

TudorAndreiTicudean:
> The driver naming here of `ad9213` is non-standard. We always use a convention `axi-<part>-rx-hpc`. I guess this is a devicetree naming bug?

Checked the ADI Linux dev branch ad9213_check, in drivers/iio/adc/ad9213.c the device name is hardcoded as "ad9213" (line 692) and the DT compatible is "adi,ad9213". Note this branch is still in development and not yet upstream. I think this might follow the SPI device convention rather than the AXI core axi-<part>-rx-hpc naming, but I'm not sure, happy to update if needed or ask the Linux team.

Update: I asked the person that created the PR about the naming. From what I understand it should be parsed from device tree and not hardcoded.

Update 2: Person answered: The ad9213 does not use "axi_adc", it uses "iio_backend" so the naming should be correct in this case. Maybe it could be named to "ad9213_phy" to be more precise.

adi/ad9213.py
```diff
@@ -3,14 +3,35 @@
 # SPDX short identifier: ADIBSD
 
 from adi.context_manager import context_manager
-from adi.rx_tx import rx_def
+from adi.rx_tx import rx
 
 
-class ad9213(rx_def, context_manager):
-    """ AD9213 High-Speed ADC """
+class ad9213(rx, context_manager):
```
tfcollins: Why are you using the rx class directly here vs the rx_def? rx_def would give you more functionality.

tfcollins (commented):
Can to address the inheritance usage question of rx vs rx_def

PR #771, opened 2026-06-24
compat: stop double-swapping big-endian RX channel data (fixes #760)

### LINKED ISSUE #760: Big endian buffer endianness is swapped twice
**Issue type:** bug
**Summary:** When receiving buffers which require endianness swaps, the swap is wrongly done twice.
**Proposed fix:** revert #647. But first, understand why it was needed in the first place.

My test setup consists of an ADIS16-series IMU connected to a Raspberry Pi running the latest analogdevicesinc/linux rpi-6.12.y kernel, with a very basic pyadi-iio app:

```python
adis = adi.adis16475("ip:...", device_name="adis16470")

adis.rx_output_type = "raw"
adis.rx_buffer_size = 1
adis.rx_enabled_channels = ["timestamp", "accel_x", "accel_y", "accel_z"] # timestamp is little endian, others are big endian
adis.rx_annotated = True

while True:
    sample_set = adis.rx()

    timestamp = int(sample_set["timestamp"][0])
    accel_x = sample_set["accel_x"][0]
    accel_y = sample_set["accel_y"][0]
    accel_z = sample_set["accel_z"][0]

    print(timestamp, accel_x, accel_y, accel_z)
```

---

#647 introduced a feature where if a libiio channel is big endian, `_rx_buffered_data` explicitly specifies to `np.frombuffer` that the bytearray is big endian. I belive this to be a bug, because by this point, the endianness has already been converted to native by libiio.

The relevant call chain is basically the same in both libiio v0.26 and latest main:

* [adi.compat_libiio_v0_rx._rx_buffered_data()](https://github.com/analogdevicesinc/pyadi-iio/blob/fdcb09e746e416e2ba4ed79dc999d328b11b0a43/adi/compat.py#L185) / [adi.compat_libiio_v1_rx._rx_buffered_data()](https://github.com/analogdevicesinc/pyadi-iio/blob/fdcb09e746e416e2ba4ed79dc999d328b11b0a43/adi/compat.py#L59)
* libiio.Channel.read() with default `raw=False` -- [v0.26](https://github.com/analogdevicesinc/libiio/blob/v0.26/bindings/python/iio.py#L847), [main](https://github.com/analogdevicesinc/libiio/blob/be9f7de36cbdd65b0c7ce493bc465455095d479b/bindings/python/iio/__init__.py#L906)
* iio_channel_read() -- [v0.26](https://github.com/analogdevicesinc/libiio/blob/v0.26/channel.c#L642), [main](https://github.com/analogdevicesinc/libiio/blob/be9f7de36cbdd65b0c7ce493bc465455095d479b/channel.c#L627)
* iio_channel_convert() -- [v0.26](https://github.com/analogdevicesinc/libiio/blob/v0.26/channel.c#L566), [main](https://github.com/analogdevicesinc/libiio/blob/be9f7de36cbdd65b0c7ce493bc465455095d479b/channel.c#L555)
* byte_swap() if appropriate

---

This surely needs more testing, especially to understand why the change was needed in the first place -- Could this possibly be a device-specific issue?

Scottcjn:
## Summary

Fixes #760 — big-endian RX channel data was being byteswapped twice, corrupting every sample on real big-endian channels (e.g. the accel channels of an ADIS16470 on a little-endian host).

## Root cause

`#647` added, in both the libiio v0 and v1 `_rx_buffered_data` paths:

```python
fmt = ">" + fmt if df.is_be else fmt
```

This forces `np.frombuffer` to interpret the bytes as big-endian whenever the channel's hardware layout is big-endian. The problem is that the bytes handed to `np.frombuffer` are **not raw** — `Channel.read()` (called just above, with the default `raw=False`) runs libiio's `iio_channel_convert()`, which already normalizes raw sample bytes to **host-native** byte order. The `# Do local type conversion` comment on that `read()` call documents exactly this.

So on a little-endian host with a big-endian channel:

1. libiio converts BE hardware bytes → LE host order (correct, native).
2. `dtype=">u2"` then tells numpy to byteswap *again* → wrong value.

`df.is_be` describes the channel's **silicon** layout, not the layout of the buffer you're actually handed after conversion — conflating the two is the bug.

## Why revert is the right fix

Per the discussion in #760, pyadi-iio handled channel endianness correctly **before** #647. The case #647 was written for (PQMON) turned out to be a firmware bug — the firmware reported the wrong channel endianness (analogdevicesinc/no-OS#3150) — not something pyadi should compensate for. Compensating in pyadi breaks every correctly-behaving big-endian device.

This PR reverts the two added lines, restoring native-endian interpretation of the already-converted buffer.

## Validation

- Data-path verified: `Channel.read()` → `iio_channel_convert()` → host-native, so the numpy dtype must be native for all channels (the reporter's mixed timestamp-LE / accel-BE set all come back native).
- The change is a literal 2-line revert of #647; no other behavior is touched.
- Hardware confirmation on the reporter's ADIS16470 setup would be the final sign-off — happy to iterate if any device path still needs the BE flag.

### DIFF
--- a/adi/compat.py
+++ b/adi/compat.py
@@ -60,7 +60,6 @@ class compat_libiio_v1_rx:
             # create format strings
             df = chan.data_format
             fmt = ("i" if df.is_signed is True else "u") + str(df.length // 8)
-            fmt = ">" + fmt if df.is_be else fmt
             data_channel_interleaved.append(np.frombuffer(bytearray_data, dtype=fmt))
 
         return data_channel_interleaved
@@ -186,7 +185,6 @@ class compat_libiio_v0_rx:
             # create format strings
             df = chan.data_format
             fmt = ("i" if df.is_signed is True else "u") + str(df.length // 8)
-            fmt = ">" + fmt if df.is_be else fmt
             data_channel_interleaved.append(np.frombuffer(bytearray_data, dtype=fmt))
 
         return data_channel_interleaved


tfcollins:
Thanks for the contribute. Please sign your commits so we can merge

tfcollins:
Fixed here https://github.com/analogdevicesinc/pyadi-iio/pull/769

PR #793, opened 2026-08-06
AD5686 update

rodrigo455:
# Description

Adds new supported devices, gain control and triggered buffer features.

References:

* https://github.com/analogdevicesinc/linux/pull/3460

Triggered Buffer output in the notebook example (updating 2 channels with hrtimer trigger):

image

## Type of change

- [ ] Bug fix
- [ ] New device class interface
- [x] New feature on an existing class
- [ ] Breaking change
- [ ] Documentation only
- [ ] Test or CI only

# How has this been tested?

Most tests use the emulated context and mock interfaces. Example code was used with real hardware.

- Hardware / emulation:
- Commands run:
    - `python -m pytest -v test/test_ad5686.py`
    - `python -m pytest -v test/test_trigger.py`
- Result:

**Test configuration**
* Kernel / libiio version: 6.12
* OS: Raspbian
* FPGA carrier (if applicable):

# Documentation

- [ ] No documentation change needed
- [x] Added a new device page and linked it from the relevant toctree

# New device class interfaces

- [x] Class extends one of the common base classes in `adi.device_base` (`rx_chan_comp`, `tx_chan_comp`, or a `_no_buff` variant). If not, explain why in the description. See the [Device Base Classes](https://analogdevicesinc.github.io/pyadi-iio/dev/device_base.html) developer doc page.
- [x] `compatible_parts` lists every part number this class supports
- [x] Part numbers added to `supported_parts.md` (verify with `invoke checkparts`)
- [ ] Emulation context (xml_gen, not iio_genxml) added under `test/emu/` and referenced from a test
- [x] Tests added that run against the emulation context in CI

# Checklist

- [x] `invoke precommit` passes locally
- [x] All commits are signed off (`Signed-off-by: Name <email>`)
- [ ] No new warnings from the build or doc generation
- [x] Dependent changes in libiio, kernel, or HDL are merged and referenced


adi/ad5686.py
```diff
@@ -102,7 +140,78 @@ class ad5686(tx_chan_comp_no_buff):
...
+            ch_voltage = (
+                voltage_np if not isinstance(voltage_np, list) else voltage_np[0]
+            )
+            data_np = self._tx_voltage_format(scale, ch_voltage)
+        else:
+            data_np = []
+            for ch_voltage in voltage_np:
+                data_np.append(self._tx_voltage_format(scale, ch_voltage))
+
+        self.tx(data_np)
```
tfcollins: Can we put this with rx_tx.py ? There is already a function on the rx side https://analogdevicesinc.github.io/pyadi-iio/buffers/index.html#buffer-units

adi/ad5686.py
```diff
@@ -102,7 +140,78 @@ class ad5686(tx_chan_comp_no_buff):
...
+
+    def set_gain(self, value: "ad5686.gain"):
+        if len(self._scales) != 2:
```
tfcollins, line 163: need doc string

tfcollins (changes requested):
If you can hoist the tx SI function into rx_tx that would be very helpful.

Also, I know you heavily mocked things for testing but we normally rely on iio-emu for this. Is iio-emu missing some functionality?

adi/ad5686.py
rodrigo455: done. I noted that "SI" is misleading as voltage/current in IIO is measured in mV/mA. IIO uses the "processed" vs "raw" terminology, which I am not sure if it sounds better either.

